PR #11, opened 2019-06-24
feat(sky): add ground and sky

mostaphaRoudsari:
This is the minimum needed to get a sky and ground for daylight factor studies. The rest of the skies will need implementing radiance commands which will come right after this.


honeybee_radiance/lightsource/sky/certainirradiance.py
```diff
@@ -0,0 +1,160 @@
...
+
+    @irradiance.setter
+    def irradiance(self, irradiance):
+        self._irradiance = typing.float_positive(irradiance) \
+            if irradiance is not None else 558.659
+
+    @property
+    def illuminance(self):
+        """Sky illuminance value."""
+        return round(self._irradiance * 179.0, 2)
```
chriswmackey: This 179 factor is confusing (I can see why Greg gets a lot of question about it).  If `illuminance` is going to be a property here, I think you should explain where it comes from in the docstring.  Or perhaps you can tell people, if they actually need to estimate illuminance values from irradince, they should not use this property but should be using [Perez's luminous efficacy model that is implemented in ladybug core](https://github.com/ladybug-tools/ladybug/blob/master/ladybug/skymodel.py#L236-L352)

honeybee_radiance/lightsource/sky/hemisphere.py
```diff
@@ -0,0 +1,117 @@
...
+import os
+
+
+class Hemisphere(object):
+    """Radiance sky hemisphere."""
+
+    def __init__(self):
+        """Create sky hemisphere.
+
+        Sky hemisphere is an upward facing glow source. Here is the default sky
```
chriswmackey: If it's a sky hemisphere, doesn't the glow source point downward to illuminate the ground?

honeybee_radiance/lightsource/sky/hemisphere.py
```diff
@@ -0,0 +1,117 @@
...
+        """Sky hemisphere emittance values for blue channel (Default: 1.0)."""
+        return self._b_emittance
+
+    @b_emittance.setter
+    def b_emittance(self, value):
+        self._b_emittance = typing.float_in_range(value)
+
+    @classmethod
+    def from_dict(cls, input_dict):
+        """Create sky_hemisphere from_dict."""
```
chriswmackey: You might want to put a sample dictionary in the Docstring here.

honeybee_radiance/lightsource/sky/certainirradiance.py
```diff
@@ -0,0 +1,160 @@
...
+        return False
+
+    @property
+    def is_climate_based(self):
+        """Return True if the sky is created based on values from weather file."""
+        return False
+
+    @classmethod
+    def from_dict(cls, input_dict):
+        """Create the sky from a dictionary."""
```
chriswmackey: You might want to put a sample dictionary in the Docstring here.

honeybee_radiance/lightsource/ground.py
```diff
@@ -0,0 +1,124 @@
...
+        """Ground emittance values for blue channel (Default: 1.0)."""
+        return self._b_emittance
+
+    @b_emittance.setter
+    def b_emittance(self, value):
+        self._b_emittance = typing.float_in_range(value)
+
+    @classmethod
+    def from_dict(cls, input_dict):
+        """Create ground from_dict."""
```
chriswmackey: You might want to put a sample dictionary in the Docstring here.

honeybee_radiance/lightsource/sky/certainirradiance.py
```diff
@@ -0,0 +1,160 @@
+"""Generate sky with certain irradiance."""
+from __future__ import division
+from ._skybase import Sky
+from .hemisphere import Hemisphere
+from honeybee_radiance.lightsource.ground import Ground
+import honeybee_radiance.typing as typing
```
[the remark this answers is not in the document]
mostaphaRoudsari: Sure. I'll do this right before merging develop to master.

honeybee_radiance/lightsource/sky/certainirradiance.py
```diff
@@ -0,0 +1,160 @@
...
+    __slots__ = (
+        '_irradiance', '_ground', '_ground_reflectance',
+        '_ground_hemisphere', '_sky_hemisphere'
+    )
+
+    def __init__(self, irradiance=558.659, ground_reflectance=0.2):
+        """Create sky with certain irradiance.
+
+        Default value is set to 558.659 which corresponds to a sky with 100,000 lux
+        horizontal illuminance.
```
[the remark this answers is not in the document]
mostaphaRoudsari: Andy has replied to a similar question here:  https://discourse.radiance-online.org/t/why-global-horizontal-illuminance-is-smaller-than-global-horizontal-radiation-multiplied-by-179-in-epw-file/4482/3

honeybee_radiance/lightsource/sky/certainirradiance.py
mostaphaRoudsari: Yes. This is specific to Radiance. You can read more about this on the Radiance forum including [this one](https://discourse.radiance-online.org/t/coefficient-179/547) and [this one](https://discourse.radiance-online.org/t/luminous-efficacy/1400).

> Or perhaps you can tell people, if they actually need to estimate illuminance values from irradince, they should not use this property but should be using Perez's luminous efficacy model that is implemented in ladybug core

Not sure if we want to make it complicated for people. I don't know why someone would use this method which is for a cloudy sky with a certain illuminance level to estimate illuminance value from irradiance at the first point.

honeybee_radiance/lightsource/sky/hemisphere.py
mostaphaRoudsari: Good idea. Will do.

honeybee_radiance/lightsource/ground.py
```diff
@@ -0,0 +1,124 @@
...
+"""
+import honeybee_radiance.typing as typing
+import ladybug.futil as futil
+import os
+
+
+class Ground(object):
+    """Radiance ground."""
+
+    def __init__(self):
```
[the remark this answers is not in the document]
mostaphaRoudsari: Yes. It is intentional. I wanted to make it harder for people to change it and in most cases you should not change this value at all but I can see specific cases when can be helpful to have the option.

honeybee_radiance/lightsource/sky/certainirradiance.py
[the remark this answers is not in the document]
mostaphaRoudsari: Also see this page 357 in this document:  https://www.radiance-online.org/community/workshops/2003-berkeley/presentations/Mardaljevic/rwr_ch6.pdf

image

honeybee_radiance/lightsource/sky/hemisphere.py
mostaphaRoudsari: You are right and this is confusing. I double checked to ensure I'm not making a mistake here. The source is upward facing in the definition.

sky_glow source sky
0
0
4 __0 0 1__ 180

But I'm not sure how this works in `skyfunc`. Will update the docstring to ensure it doesn't confuse other users.

honeybee_radiance/lightsource/sky/certainirradiance.py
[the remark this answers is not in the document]
chriswmackey: Thank you for posting the links but I don't think I need to see them as much as future people reading the docstring who do not understand Radiance's built in conversion factor.  Why not just paste that description of how Radiance deals with luminous efficacy into the docstring?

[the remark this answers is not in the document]
mostaphaRoudsari: good idea. added it to class docstring with a links to forum discussions.

PR #13, opened 2019-06-29
Add ModifierSet

### LINKED ISSUE #7: add radiance command options
Add what is needed for `rtrace` and that will be enough for now.

### LINKED ISSUE #8: start adding radiance commands
I need to add

- [ ] command base
- [ ] rtrace

### LINKED ISSUE #9: update view properties to RadianceOptions
See here:

https://github.com/ladybug-tools/honeybee-radiance/pull/6#discussion_r294825055

### LINKED ISSUE #12: Switch honeybee_radiance to use the typing module in honeybee_core
Honeybee-radiance should be using the same typing module as honeybee-energy and honeybee-core found here:
https://github.com/ladybug-tools/honeybee-core/blob/master/honeybee/typing.py

mostaphaRoudsari:
This PR:

<s>1. adds base classes for commands and options</s>
<s>2. adds oconv, rtrace, rcontrib and rcalc</s>
3. replaces `honeybee_radiance.typing` with `honeybee.typing`
4. updates view class to use `RadianceOptions`
5. adds modifier sets

<s>Resolves #7</s>
<s>Resolves #8</s>
Resolves #9 
Resolves #12


honeybee_radiance/command/options/oconv.py
```diff
@@ -0,0 +1,115 @@
...
+    def on_setattr(self):
+        assert not (self.b.is_set and self.i.is_set), \
+            'The -b and -i options are mutually exclusive.'
+
+    @property
+    def i(self):
+        """input octree."""
+        return self._i
+
+    @i.setter
```
chriswmackey: If you aren't performing any checks on the setters of this class do you actually need them as setters?  Can't they just be attributes?  I am also a bit unclear on what type of object they are supposed to be. Is is a string?

honeybee_radiance/command/options/oconv.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        return self._b
+
+    @b.setter
+    def b(self, value):
+        self._b.value = value
+
+    @property
+    def n(self):
+        """maximum surface set size for each voxel.
```
chriswmackey: For consistency sake with documentation, you should probably capitalize Maximum.  The same goes for a bunch of other docstrings on this object.

honeybee_radiance/command/options/rcalc.py
```diff
@@ -0,0 +1,200 @@
...
+        self._e = StringOption('e', 'expression', pattern_out='"%s"')
+        self._on_setattr_check = True
+
+    def on_setattr(self):
+        if self.p.is_set and not self.i.is_set:
+            warnings.warn('rcalc: -p has no effect unless -i is also specified.')
+
+    @property
+    def tS(self):
+        """alternative tab character."""
```
chriswmackey: Docstrings are not capitalized throughout this file.

honeybee_radiance/command/options/rcontrib.py
```diff
@@ -0,0 +1,234 @@
...
+
+# TODO: Add input pattern check for -p
+class RcontribOptions(RtraceOptions):
+    """rcontrib command options.
+
+    [ -n nprocs ][ -V ][ -c count ][ -fo | -r ][ -e expr ][ -f source ][ -o ospec ]
+    [ -p p1=V1,p2=V2][ -b binv ][ -bn nbins ] { -m mod | -M file } [ $EVAR ] [ @file ]
+    [ rtrace options ]
+    
+    NOTE:
```
chriswmackey: Again, I am not sure if this is supposed to be all caps.

chriswmackey: Also, I see in some places you use `NOTE:` and in others you use `Also see:`  You might just want to be consistent about this.

honeybee_radiance/command/options/rtrace.py
```diff
@@ -0,0 +1,890 @@
...
+    """rtrace command options.
+
+    Also see: https://floyd.lbl.gov/radiance/man_html/rtrace.1.html
+    """
+
+    __slots__ = (
+        '_n', '_x', '_y', '_ld', '_h', '_fio', '_o', '_w', '_i', '_u', '_bv', '_dt',
+        '_dc', '_dj', '_ds', '_dr', '_dp', '_dv', '_ss', '_st', '_av', '_aw', '_ab',
+        '_aa', '_ar', '_ad', '_as', '_me', '_ma', '_mg', '_ms', '_lr', '_lw', '_I',
+        '_e', '_te', '_tE', '_ti', '_tI', '_af', '_ae', '_ai', '_aE', '_aI'
```
chriswmackey: I think rtrace officially wins for the largest `__slots__`

honeybee_radiance/command/options/rtrace.py
```diff
@@ -0,0 +1,890 @@
...
+        self._tI = FileOption('tI', 'trace included modifiers file')
+        self._af = StringOption('af', 'ambient file')
+        self._ae = StringOption('ae', 'ambient excluded modifier')
+        self._aE = FileOption('aE', 'ambient excluded modifiers file')
+        self._ai = StringOption('ai', 'ambient included modifier')
+        self._aI = FileOption('aI', 'ambient included modifiers file')
+        self._e = FileOption('e', 'error file')
+        self._on_setattr_check = True
+
+    def on_setattr(self):
```
chriswmackey: It does seem a bit odd to me that the methods going by this name are not hidden.  Otherwise, if you are going to expose them, maybe they should get docstrings.

honeybee_radiance/command/options/rtrace.py
```diff
@@ -0,0 +1,890 @@
...
+                else:
+                    print('-lw ({}) is already <= {}.'.format(self.lw, suggested_lw))
+
+    @property
+    def n(self):
+        """number of rendering processes - default: 1
+        
+        Execute in parallel on nproc local processes.
+        
+        NOTE:
```
chriswmackey: Another all-caps NOTE.  Feel free to resolve all of these if the all-caps format is acceptable by sphinx.

honeybee_radiance/command/rcontrib.py
```diff
@@ -0,0 +1,38 @@
...
+    Rcontrib computes ray coefficients for objects whose modifiers are named in one or
+    more -m settings. These modifiers are usually materials associated with light sources
+    or sky domes, and must directly modify some geometric primitives to be considered in
+    the output. A modifier list may also be read from a file using the -M option. The
+    RAYPATH environment variable determines directories to search for this file. (No
+    search takes place if a file name begins with a '.', '/' or '~' character.).
+
+    See RcontribOptions class for more information.
+
+    NOTE:
```
chriswmackey: NOTE: There is a NOTE here.

honeybee_radiance/exception.py
```diff
@@ -0,0 +1,61 @@
+"""Honeybee radiance exceptions."""
```
chriswmackey: Very cool thing you did here.  We might consider something similar for all of the exceptions that can get raised from the checking methods in honeybee-core.

honeybee_radiance/lightsource/sky/certainirradiance.py
```diff
@@ -22,6 +22,17 @@ class CertainIrradiance(Sky):
     It also includes ground glow source. Ground reflectance is set to %20 by default
     which is gensky's default value. Use `ground_reflectance` property to adjust this
     value.
+
+    NOTE:
+
+    The conversion factor in the Radiance system for luminous efficacy is fixed at
+    KR= 179 lumens/watt (lm/w). This should not be confused with the more usual
```
chriswmackey: Thank you for adding this.

honeybee_radiance/view.py
```diff
@@ -31,30 +33,24 @@ class View(object):
...
-
-    def __init__(self, name, origin=None, direction=None, up_vector=None, type=0,
-                 h_size=60, v_size=60, shift=0, lift=0):
+    def __init__(self, name, position=None, direction=None, up_vector=None, type='v',
+                 h_size=60, v_size=60, shift=None, lift=None):
         u"""Create a view.
 
         Arg:
-            origin: Set the view origin location (-vp) to (x, y, z). This is the focal
+            position: Set the view position location (-vp) to (x, y, z). This is the focal
```
chriswmackey: `position location`  sounds a bit redundant.  I think you can just say `position`

honeybee_radiance/command/options/oconv.py
chriswmackey: And for nearly all of the docstrings copy/pasted from radiance for these parameters.

honeybee_radiance/command/_command.py
```diff
@@ -0,0 +1,172 @@
...
+        if self.output and self.pipe_to:
+            warnings.warn(
+                '%s: "output" and "pipe_to" are both set for this command.'
+                ' The "output" argument  will be ignored.'
+                % self.__class__.__name__.lower()
+            )
+
+    def run(self, env=None, cwd=None, wait=True):
+        """Run command as a subprocess."""
+        NotImplementedError()
```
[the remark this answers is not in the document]
mostaphaRoudsari: This was a place-holder.

honeybee_radiance/command/oconv.py
```diff
@@ -0,0 +1,87 @@
...
+            value = OconvOptions()
+
+        if not isinstance(value, OconvOptions):
+            raise ValueError('Expected OconvOptions not {}'.format(type(value)))
+
+        self._options = value
+
+    @property
+    def inputs(self):
+        """Octree file.
```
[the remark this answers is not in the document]
mostaphaRoudsari: This will used in commands. thank you for pointing it out.

honeybee_radiance/command/options/_optutil.py
```diff
@@ -0,0 +1,171 @@
...
+    '    """{2} command options."""\n\n' \
+    '    __slots__ = ({3})\n\n' \
+    '    def __init__(self):\n' \
+    '        """{2} command options."""\n'
+    '        OptionCollection.__init__(self)\n' \
+
+    # run the command a parse the output
+    def get_defaults(command):
+        data = subprocess.check_output([command, '-defaults'])
+        # only works in python 3
```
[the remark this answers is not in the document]
mostaphaRoudsari: Yes. This will only be used by developers to generate a draft for Radiance command options.

honeybee_radiance/command/options/oconv.py
mostaphaRoudsari: The check is done in line 18. We still need a setter to ensure the check happens.

honeybee_radiance/exception.py
mostaphaRoudsari: That can be. Custom exceptions are definitely helpful but I think we should just add them if the normal exceptions are not serving our purpose. Otherwise we can end up with one more thing to maintain and for no good reason.

honeybee_radiance/command/options/rtrace.py
mostaphaRoudsari: added documentation and also made it hidden.

mostaphaRoudsari:
@chriswmackey, I addressed your previous comments for this PR but I also made a major change that needs your review.

1. Command and options classes are now moved to a separate repository and are removed from this repostiory.
2. I added ModifierSet class which closely follows your implementation of ConstructionSet. I realized you can save several lines of code in Sets by using `__slots__` from the base class. Let me know if you want me to make similar changes to Energy ConstructionSet and send a PR.

https://github.com/ladybug-tools/honeybee-radiance/pull/13/files#diff-75f2991509aac6c4b160bb084f6022c9R175

chriswmackey:
@mostaphaRoudsari ,
Sorry that I didn't check my PR review requests until now.  I'm pretty tired at the moment so I think the best course of action is to do a full review first thing tomorrow.

To give my quick thoughts right now, I am really happy that you have implemented a `MaterialSet` and I am also really happy to see that it uses the `GlassDoor` aperture type.  In light of this, I really think it's worth it to find a way to ensure the aperture types are consistent across the extensions.  It will just make integration across the extensions a lot easier to manage.

If you think that there isn't a case where someone would want to treat an operable window differently from a fixed window in Radiance simulations, then maybe I can just move all properties related to opening windows into the ApertureEnergyProperties rather than having them implemented as an Aperture type.  Let me know if you have any immediate thoughts on this proposal.  Otherwise we can just discuss it on our next call.

honeybee_radiance/modifierset.py
```diff
@@ -0,0 +1,702 @@
...
+
+@lockable
+class ApertureSet(_BaseSet):
+    """Set containing all radiance modifiers needed to for a radiance model's Apertures.
+
+    Properties:
+        exterior_modifier
+        interior_modifier
+        skylight_modifier
+        glass_door_modifier
```
chriswmackey: The more that I think about it what I would be doing if I were in practice, the more that I realize that I would find myself writing scripts to account for insect screens in operable windows such that they could be differentiated from fixed windows, which could be completely clear of the insect screens.  Just from looking at insect screens, it is clear that they block out at least as much light as another pane of glass would and, if we are already going through the trouble to make a distinction between an exterior double pane assembly and an interior single pane one, the insect screens seem to be just as relevant of an issue.

I also realize that this is a concern that is not relevant everywhere in the world since insect screens are more common in places with lots of flying insects, like the eastern US or China (I know that people mostly do not need them in Europe).

But I just want to go on the record saying that, if `operable_window_modifier` were included here in the `ModifierSet`, I would be one of the people using it. I don't know if I can really speak for others here so, I understand if you want to ask more people in the daylighting community before making a decision @mostaphaRoudsari .

honeybee_radiance/modifierset.py
```diff
@@ -0,0 +1,702 @@
...
+            return self._get_modifier_from_set(self.floor_set, boundary_condition)
+        elif face_type == 'RoofCeiling':
+            return self._get_modifier_from_set(self.roof_ceiling_set, boundary_condition)
+        elif face_type == 'AirWall':
+            return air_wall
+        else:
+            raise NotImplementedError(
+                'Face type {} is not recognized for ModifierSet'.format(face_type))
+
+    def get_aperture_modifier(self, boundary_condition, aperture_type, parent_face_type):
```
chriswmackey: I am glad to see that we are using the same logic here as `honeybee-energy`.  I also see the acknowledgement of the `OperableWindow` aperture type, making me more confident that it should be moved to the core.

honeybee_radiance/modifierset.py
```diff
@@ -0,0 +1,702 @@
...
+        Args:
+            data: Dictionary describing the ModifierSet.
+        """
+        assert data['type'] == 'ModifierSet', \
+            'Expected ModifierSet. Got {}.'.format(data['type'])
+
+        # TODO: build each modifier from JSON
+        # TODO: put each set together
+        # NOTE: An alternative solution is to implement from_dict to sets
+        # and use that to recreate the class.
```
chriswmackey: I would say that you should feel free to go for this alternative here.  The only reason why it did not make sense in honeybee-energy is because you often have a lot of materials shared between all of the constructions across the set.  You don't have the same restriction here and, while there might be a few duplicated modifiers across the whole set, I don't think that you are saving enough duplication to merit a solution like that in `honeybee-energy`.

honeybee_radiance/modifierset.py
```diff
@@ -0,0 +1,702 @@
...
+    def interior_modifier(self, value):
+        self._interior_modifier = self._validate_modifier(value)
+
+    @property
+    def modifiers(self):
+        """List of all modifiers contained within the set."""
+        return [getattr(self, attr[1:]) for attr in self._slots]
+
+    @property
+    def modified_modifiers(self):
```
chriswmackey: Hm. I realize that I may not have thought through all of the implications of this naming choice.  I don't really have a better solution to offer at this time but, if you come up with a better convention that `modified`, I am happy to implement it on the honeybee-energy side as well.

honeybee_radiance/lib/modifiers.py
```diff
@@ -0,0 +1,13 @@
+"""Library of default modifiers for honeybee-radiance."""
+
+from honeybee_radiance.primitive.material import Plastic, Glass
+
+
+generic_exterior_wall = Plastic.from_single_reflectance('generic_ext_wall_0.35', 0.35)
+generic_interior_wall = Plastic.from_single_reflectance('generic_int_wall_0.50', 0.5)
+generic_ceiling = Plastic.from_single_reflectance('generic_ceiling_0.80', 0.8)
+generic_roof = Plastic.from_single_reflectance('generic_roof_0.20', 0.2)
```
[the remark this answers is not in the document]
mostaphaRoudsari: This one is tricky because as you noted in a single face model ceiling and roof will be the same face. I wish there was a solution to pick two different materials for two different sides of a face but for now I take your suggestion and change it to 0.7

honeybee_radiance/lib/modifiers.py
```diff
@@ -0,0 +1,13 @@
...
+generic_floor = Plastic.from_single_reflectance('generic_floor_0.20', 0.2)
+generic_exterior_glass = Glass.from_single_transmittance('generic_ext_glass_0.60', 0.6)
+generic_interior_glass = Glass.from_single_transmittance('generic_int_glass_0.60', 0.6)
```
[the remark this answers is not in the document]
mostaphaRoudsari: Good point. I was trying to be conservative here. Will change this to 0.8 for interior.

honeybee_radiance/lib/modifiers.py
```diff
@@ -0,0 +1,13 @@
...
+air_wall = Glass.from_single_transmissivity('air_wall', 1.0)
```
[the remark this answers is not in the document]
mostaphaRoudsari: I will but that needs some extra work. I need to make them lockable first. There is an open issue to address this: https://github.com/ladybug-tools/honeybee-radiance/issues/14

honeybee_radiance/modifierset.py
```diff
@@ -0,0 +1,702 @@
...
+    """
+    __slots__ = ()
+
+
+@lockable
+class ApertureSet(_BaseSet):
+    """Set containing all radiance modifiers needed to for a radiance model's Apertures.
+
+    Properties:
+        exterior_modifier
```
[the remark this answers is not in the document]
mostaphaRoudsari: Actually now that you posted this it got me thinking that and Aperture by definition cannot be interior so the logical move is to remove the interior one but then where do we address interior windows?

The window from aperture distinction has started to be an issue. Not sure what is the best solution here but what I think should not change is that an aperture is an exterior window, skylight or glass door.

>  But I also realize that, as a developer, you are probably in love with the beauty of inherited code as I would be had I written it.

[?] - not sure if I follow all the love and feeling here! 😕 I think once we can sort out the distinction between interior window and aperture this will be sorted out by default. Maybe I should double check how you handled it for `honeybee-energy`.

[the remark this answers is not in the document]
mostaphaRoudsari: Maybe it should be a WindowSet instead of ApertureSet?

honeybee_radiance/modifierset.py
mostaphaRoudsari: Let me re-think the categories for aperture / windows.

To me, the current logic that you are trying to come up with is backwards! If you are concerned about an extra screen just make the glass darker for those windows. I don't think you need an operable window category for that but I see your point.

Operable window can be interesting as the light transmittance will changed based on window being open or closed (or have insect screens or not). The question is if this should be addressed here or should be modeled as a Dynamic Window? I think the second one gives full control on how one wants to model the glass.

Also this is a default library for getting a minimum ready-to-go model. I think some of the complexity that you are talking about is out of the scope for default material set.

honeybee_radiance/modifierset.py
mostaphaRoudsari: let's chat about this. I think you have a point and the operable window type started to be a challenge. I think it will be even more of a challenge when we want to also handle CFD models. If you think moving it to honeybee-core will be helpful then I agree that we should do it.

And yes. I tried to keep the API consistent with `honeybee-energy` between the two so we can reuse it between the libraries.

honeybee_radiance/modifierset.py
mostaphaRoudsari: I didn't want to suggest a major change in the middle of your current development and I as I said I tried to keep the methods the exact same name. I personally think we can use `set` instead of `modified`. These are the modifiers / constructions that are __set__ by user. In this case it will be `set_modifiers` or `modifiers_set_by_user`.

I should also say that it is fine to have it as is. One can read the docstring and that should tell them what this is.

PR #16, opened 2019-08-29
fix(modifier-set): update modifier-set

mostaphaRoudsari:
Updated modifier-set to match the latest changes in `honeybee-core`.

### DIFF
--- a/honeybee_radiance/lib/modifiers.py
+++ b/honeybee_radiance/lib/modifiers.py
@@ -1,17 +1,39 @@
 """Library of default modifiers for honeybee-radiance.
 
-Default values are generic values to set the initial visible reflectance and
-transmittance values in your model. There is no guarantee that these values exactly matches your model.
+Default values are generic values to set the initial visible / solar reflectance and
+transmittance values in your model. There is no guarantee that these values exactly
+match what you are trying to model.
 """
 
 from honeybee_radiance.primitive.material import Plastic, Glass
 
-
-generic_exterior_wall = Plastic.from_single_reflectance('generic_ext_wall_0.35', 0.35)
+# interior modifiers - visible
+generic_floor = Plastic.from_single_reflectance('generic_floor_0.20', 0.2)
 generic_interior_wall = Plastic.from_single_reflectance('generic_int_wall_0.50', 0.5)
 generic_ceiling = Plastic.from_single_reflectance('generic_ceiling_0.80', 0.8)
-generic_roof = Plastic.from_single_reflectance('generic_roof_0.70', 0.7)
-generic_floor = Plastic.from_single_reflectance('generic_floor_0.20', 0.2)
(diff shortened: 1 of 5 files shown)


honeybee_radiance/lib/modifiers.py
```diff
@@ -1,17 +1,39 @@
 """Library of default modifiers for honeybee-radiance.
 
-Default values are generic values to set the initial visible reflectance and
-transmittance values in your model. There is no guarantee that these values exactly matches your model.
+Default values are generic values to set the initial visible / solar reflectance and
+transmittance values in your model. There is no guarantee that these values exactly
+matches your model.
```
chriswmackey: "match what you are trying to model?"
I am pretty sure that the match the model this is built/exported.

honeybee_radiance/modifierset.py
```diff
@@ -29,13 +31,19 @@
...
+        'skylight': generic_exterior_glass
     },
     'DoorSet': {
-        'exterior': generic_exterior_wall,
-        'interior': generic_interior_wall,
+        'exterior': generic_opaque_door,
+        'interior': generic_opaque_door,
+        'exterior_glass': generic_exterior_glass_door,
+        'interior_glass': generic_interior_glass_door,
         'overhead': generic_interior_wall
```
chriswmackey: Why not use generic_opaque_door here?

honeybee_radiance/modifierset.py
```diff
@@ -218,34 +225,56 @@ class ApertureSet(_BaseSet):
...
-    def __init__(self, exterior_modifier=None, interior_modifier=None,
-                 skylight_modifier=None, glass_door_modifier=None):
+    def __init__(self, window_modifier=None, interior_modifier=None,
+                 skylight_modifier=None, operable_modifier=None):
         """Initialize aperture set.
 
         Args:
-            exterior_modifier: A window modifier object for apertures
+            window_modifier: A window modifier object for apertures
                 with an Outdoors boundary condition.
```
chriswmackey: An Outdoors boundary condition and Wall parent Face.

honeybee_radiance/modifierset.py
```diff
@@ -218,34 +225,56 @@ class ApertureSet(_BaseSet):
...
             interior_modifier: A window modifier object for apertures
                 with a Surface boundary condition.
             skylight_modifier: : A window modifier object for apertures with a
                 Outdoors boundary condition, and RoofCeiling or Floor face type for
                 their parent face.
-            glass_door_modifier: A window modifier object for apertures
+            operable_modifier: A window modifier object for apertures
                 with an Outdoors boundary condition and GlassDoor aperture type.
```
chriswmackey: With an Outdoors boundary condition and a True is_operable property.

mostaphaRoudsari:
Hey Chris! Thank you for the comments. I addressed all of them.

PR #20, opened 2019-10-25
ci(docs): separate building docs and creating PyPI distribution

mostaphaRoudsari:
I noticed the the online documentation for honeybee-radiance has been overwritten after the last
commit which was a chore(). This change should fix that issue from now on.

### DIFF
--- a/.releaserc.json
+++ b/.releaserc.json
@@ -6,7 +6,7 @@
         [
             "@semantic-release/exec",
             {
-                "publishCmd": "bash deploy.sh ${nextRelease.version}"
+                "publishCmd": "bash deploy.sh"
             }
         ]
     ]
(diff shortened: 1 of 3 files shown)


AntoineDao:
This change will make it build docs and overwrite them every time you commit to master. This is what you want yes?

mostaphaRoudsari:
yes! otherwise all the non-release commits will end up overwriting the previous docs with an empty repository.

and now I see that the recent changes in honeybee-core writer has made a breaking change. I can double check that and send a separate PR.

AntoineDao:
It might be worth having a chat about read the docs again some time soon. It really makes the whole docs deployment/management/versioning an absolute breeze compared to these hack CI jobs...

PR #22, opened 2019-12-03
Docker

AntoineDao:
This PR sets up this repository to deploy a newly tagged docker container each time a release is made.

Before it is accepted @mostaphaRoudsari will need to add the `DOCKER_USERNAME` and `DOCKER_PASSWORD` creds to secret env vars in the Travis config.

### DIFF
--- a/.releaserc.json
+++ b/.releaserc.json
@@ -6,7 +6,7 @@
         [
             "@semantic-release/exec",
             {
-                "publishCmd": "bash deploy.sh"
+                "publishCmd": "bash deploy.sh ${nextRelease.version}"
             }
         ]
     ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6
+FROM python:3.7
 
 WORKDIR /tmp/
 RUN curl -L https://github.com/NREL/Radiance/releases/download/5.2/radiance-5.2.dd0f8e38a7-Linux.tar.gz | tar xz \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM python:3.6
+
+WORKDIR /tmp/
+RUN curl -L https://github.com/NREL/Radiance/releases/download/5.2/radiance-5.2.dd0f8e38a7-Linux.tar.gz | tar xz
+RUN mv `ls`/usr/local/radiance/bin/* /usr/local/bin
+RUN mv `ls`/usr/local/radiance/lib/* /usr/local/lib
+RUN rm -rf `ls`
+
+WORKDIR /usr/app/
+# COPY requirements.txt requirements.txt
+# RUN pip install -r requirements.txt
+COPY . .
+RUN pip install -e .[cli]
+ENTRYPOINT [ "honeybee-radiance" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,4 +10,3 @@ WORKDIR /usr/app/
 
 COPY . .
 RUN pip install .[cli]
-ENTRYPOINT [ "honeybee-radiance" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,13 @@
 FROM python:3.6
 
 WORKDIR /tmp/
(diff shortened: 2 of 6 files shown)


.releaserc.json
```diff
@@ -6,7 +6,7 @@
         [
             "@semantic-release/exec",
             {
-                "publishCmd": "bash deploy.sh"
+                "publishCmd": "bash deploy.sh ${nextRelease.version}"
```
mostaphaRoudsari: Good one! How do you pass `{nextRelease.version}` here?

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM python:3.6
+
+WORKDIR /tmp/
+RUN curl -L https://github.com/NREL/Radiance/releases/download/5.2/radiance-5.2.dd0f8e38a7-Linux.tar.gz | tar xz
+RUN mv `ls`/usr/local/radiance/bin/* /usr/local/bin
+RUN mv `ls`/usr/local/radiance/lib/* /usr/local/lib
+RUN rm -rf `ls`
```
mostaphaRoudsari: I think we should merge these 3 lines into 1 and use &&.

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+WORKDIR /usr/app/
+# COPY requirements.txt requirements.txt
+# RUN pip install -r requirements.txt
+COPY . .
+RUN pip install -e .[cli]
```
mostaphaRoudsari: Do we really need `-e` here. Are you using the editable version to be faster?

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+ENTRYPOINT [ "honeybee-radiance" ]
```
mostaphaRoudsari: This might limit the use of command line since you might want to use `honeybee` commands insider this docker image too. I suggest to change it to "honeybee" or does not include an entry point. With "honeybee" as entrypoint one can use `honeybee radiance` to access honeybee-radiance commands.

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+WORKDIR /tmp/
+RUN curl -L https://github.com/NREL/Radiance/releases/download/5.2/radiance-5.2.dd0f8e38a7-Linux.tar.gz | tar xz
+RUN mv `ls`/usr/local/radiance/bin/* /usr/local/bin
+RUN mv `ls`/usr/local/radiance/lib/* /usr/local/lib
+RUN rm -rf `ls`
+
+WORKDIR /usr/app/
+# COPY requirements.txt requirements.txt
+# RUN pip install -r requirements.txt
```
mostaphaRoudsari: Let's remove these two lines.

setup.py
```diff
@@ -20,7 +20,7 @@
     url="https://github.com/ladybug-tools/honeybee-radiance",
     packages=setuptools.find_packages(exclude=["tests"]),
     install_requires=requirements,
-    extra_requires={
+    extras_require={
```
mostaphaRoudsari: Good catch! Thanks.

Dockerfile
AntoineDao: I reckon we should keep it that way. That way there is a clear distinction between the `honeybee-radiance` and `honeybee-core` docker containers.

mostaphaRoudsari: I like the distinction idea but thinking about it more having an entrypoint will break the ability to run the same Queenbee workflow locally. 

This will make the command to be executed in docker to change from `honeybee-radiance grid split ...` to `grid split ...`. We should either adjust Queenbee to take two commands (it actually does still have an input for local) or remove the entrypoint so we can use an identical command for local runs and running them inside a docker container. 😵

AntoineDao: Aaaaaaaaaah I see what you mean now! Hmm... I'm torn about this one actually. Let's remove the entrypoint for now to make it easy and we can review that decision at a later date.

AntoineDao:
Let me know when we're good to merge this one in.

AntoineDao:
Cool.  Will add it in just now and merge. I'm not sure how you plan on integrating `async` if we're still running tests for python 2 in this repo though... :stuck_out_tongue:

AntoineDao:
https://hub.docker.com/r/ladybugtools/honeybee-radiance/tags :tada:

PR #26, opened 2019-12-10
Couple of fixes to get workflows working

AntoineDao:
Couple of minor fixes. Tested them on workflows by creating my own docker image and running it as a new operator.

### DIFF
--- a/honeybee_radiance/cli/grid.py
+++ b/honeybee_radiance/cli/grid.py
@@ -12,6 +12,7 @@ import honeybee_radiance.sensorgrid as sensorgrid
 import os
 import logging
 import re
+import json
 
 _logger = logging.getLogger(__name__)
 
@@ -43,7 +44,8 @@ def split_grid(grid_file, count, folder, log_file):
         grid = sensorgrid.SensorGrid.from_file(grid_file)
         file_count = int(round(grid.count / count))
         files = grid.to_files(folder, file_count, mkdir=True)
-        log_file.write('\n'.join(files))
+        
+        log_file.write(json.dumps(files))
     except Exception:
         _logger.exception('Failed to split grid file.')
         sys.exit(1)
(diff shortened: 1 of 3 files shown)


honeybee_radiance/sensorgrid.py
```diff
@@ -179,8 +179,6 @@ def to_files(self, folder, count, base_name=None, mkdir=False, full_path=False):
             base_name: Optional name for base_name for sensor files (Default: self.name).
             mkdir: A boolean to indicate if the folder should be created in case it
                 doesn't exist already (Default: False).
-            full_path: A boolean to indicate if this method should return the full_path
-                to the newly created files or just the name.
 
         Returns:
             A list of filepaths to newly created files. The path can be full_path or just
```
mostaphaRoudsari: We need to update Returns docstring here

honeybee_radiance/cli/grid.py
```diff
@@ -43,7 +44,8 @@ def split_grid(grid_file, count, folder, log_file):
         grid = sensorgrid.SensorGrid.from_file(grid_file)
         file_count = int(round(grid.count / count))
         files = grid.to_files(folder, file_count, mkdir=True)
-        log_file.write('\n'.join(files))
+        
+        log_file.write(json.dumps(files))
```
mostaphaRoudsari: Curios to know why do we need json.dumps in this case?

AntoineDao: It turns out that if you load a file as an output parameter in Argo and you then want to use that parameter to loop tasks, then the parameter must be a JSON serializable array.

ie:
## BAD
```
room_0001.pts \n
room_0002.pts 
``` 

## GOOD
```
[ "room_0001.pts", "room_0002.pts" ]
```
```
[
  {
    "name": "room_0001",
    "path": "room_0001.pts"
  },
  {
    "name": "room_0002",
    "path": "room_0002.pts"
  }
]
```

mostaphaRoudsari: Got it! Thanks. This might be a better practice anyways.

PR #31, opened 2020-01-29
feat(modifier): Refactor modifier modules

chriswmackey:
This is an initial refactor of the whole library in preparation to make it an extension of honeybee-core. There are a number of edits included in it:

* Completed the ModifierSet object, including the from_dict method
* Fixed some places where native python operators (like type) were being overwritten with variable names
* Filled in a number of incorrect docstrings that were copy/pasted from other classes but never edited
* Added the term "primitive dictionary" to differentiate the generic schema of Radiance Primitives from dictionaries following a schema specific to a given primitive. I used this throughout several docstrings that I added and this should help clarify what's going on to new users.
* The use of the Point3D class in geometry primitives was unnecessary since the only purpose these primitives serve now is translation to/from Radiance. They have no need for any geometry computation and any of that should probably happen on Rooms, Faces, Apertures, Doors, Shades where we use real 3D objects like Face3D or Sphere (not just points). So I think we should leave these geometry primitive classes as something used only in the writer/reader modules for serialization. And the use of Point3D objects was actually interfering with this primary purpose by introducing an in-elegant schema for serializing points to a dictionary that's not even used within ladybug_geometry.  So we should just  keep points and vectors simple using arrays to represent them.
* Switched to using iterables for primitive values (list instead of dictionary). This uses much less memory, it's cleaner for hashing + copying, and it avoids the possibility of the user getting a KeyError when they go the check the values by index (what seems to be one of the major use cases of the property). Using lists also simplified a lot of the code but, if you are attached to the dictionaries, Mostapha, we can put them back and I'll need to add some extra code to handle cases of missing keys and the fact that the dictionary isn't always ordered). We would also have to figure out how we would want to represent such dictionaries in the JOSN schema since they technically should be their own object with their own 'type' key if they are dictionaries and not arrays.
* Added `__hash__` operator to all primitives that agrees with the `__eq__` operator
* Added `__copy__` and duplicate methods to all primitives
* Implement `__slots__` on all primitives
* Added some tests but largely refrained from editing existing tests (beyond changing the modules that were imported)


honeybee_radiance/properties.py
```diff
@@ -61,7 +62,7 @@ def modifier_dir(self):
 
     @modifier_dir.setter
     def modifier_dir(self, value):
-        if hasattr(value, 'can_be_modifier') and value.can_be_modifier:
+        if isinstance(value, Primitive) and value.is_modifier:
```
mostaphaRoudsari: Shouldn't we be able to just do `isinstance(value, Modifier)` in such cases now that it has its own type?

honeybee_radiance/primitive.py
```diff
@@ -0,0 +1,486 @@
...
+    
+    def _dup_mod_and_depend(self):
+        """Duplicate this object's modifer and its dependencies."""
+        mod = None if isinstance(self._modifier, Void) else self._modifier.duplicate()
+        dependencies = []
+        for dep in self._dependencies:
+            try:
+                dependencies.append(dep.duplicate())  # it's a prmitive like a modifer
+            except AttributeError:
+                dependencies.append(dep)  # it's an immutable object like a string
```
mostaphaRoudsari: Technically, this should never happen. All the dependencies should be some sort of modifiers. Did we have test that was breaking because of this?

honeybee_radiance/properties.py
chriswmackey: Yes, we could but there's an open question as to whether we want to allow people to assign a `Primitive` here that is of the `Primitive` base class and isn't of a specific modifier class. You can see that [the `is_modifier` property of the Primitive class](https://github.com/ladybug-tools/honeybee-radiance/pull/31/files#diff-7e8803b3f5c02bf9bde1c8b715f1ee34R309-R315) is good enough on its own to tell you whether Radiance will interpret the primitive as a modifier. But I don't know if this flexibility is useful at all and, if you think it's better to enforce that the object be of the `Modifier` subclass, I will do that.

In any case, I might leave this as-is for now since the next PR is the one where I'll be overwriting all of the extension properties.

honeybee_radiance/primitive.py
chriswmackey: No, I don't think this broke any of the tests but I remember you showing me a case where one of the dependencies was a `.cal` file and I assumed that would be coming here as a string. I'll see now if this causes any failing tests and let me know if I just wasn't understanding what a dependency could be.

chriswmackey: I can confirm that no tests failed. If you just confirm that my understanding of .cal files wasn't correct, I'll remove this.

mostaphaRoudsari: When it comes to dependency I think I only keep the modifiers that I find in the document in dependencies and ignore the rest so there will not be a case for .cal files, etc. The reason for that is I can't access the cal file even if I want to at that time and we can hope that the user puts the file in the right place at the right path.

chriswmackey: Good to know. I'll take out the try/except and merge this then. By the end of tomorrow, this library should be a full extension!

PR #33, opened 2020-01-31
feat(properties): Add extension properties for all Model obects

chriswmackey:
This commit adds extension properties for all geometry objects (Room, Face, Aperture, Door) and the Model object.

I still have to add some more tests but this code is ready for review, @mostapha .  There are still a should of things that need to be figured out:

* I was trying to implement some form of states in the Aperture properties at first but I saw the implementation of States in the older HoneybeePlus and I don't know how you see dynamic shade geometry working in a setup where you can assign shades to apertures. Given how the schema is set up, this assignment might actually interfere with how you were previously using shade states or it might end up being unnecessary/redundant unless you want to have all of the shades assigned to an Aperture on a single state.

* I don't know how you want to write out indoor vs outdoor shades (right now, the methods that translate to.rad are writing the object and all sub-objects so there's no distinction between indoor and outdoor shades) I think I will need a crash course in the radiance folder structure to understand how these should be changed.

* Should the modifiers be written into the same file as the geometry when translating a Model.to.rad?


mostaphaRoudsari:
@chriswmackey, thank you for the PR. Your questions are all very good questions. It will be easier to answer them during a call. I'm available before 11am or after 3pm.

honeybee_radiance/properties/face.py
```diff
@@ -0,0 +1,181 @@
...
+            If None, the properties will be duplicated with the same host.
+        """
+        _host = new_host or self._host
+        return FaceRadianceProperties(_host, self._modifier, self._modifier_blk)
+
+    def ToString(self):
+        return self.__repr__()
+
+    def __repr__(self):
+        return 'Face Radiance Properties:\n host: {}'.format(self.host.name)
```
mostaphaRoudsari: It looks to me that we can save a lot of very similar code by using a base class for aperture, door, shade and face. We may just need to add a new method to modifier-set to take the type name. Is there another reason that you are not using a baseclass?

honeybee_radiance/writer.py
```diff
@@ -1,22 +1,177 @@
...
+    """
+    room_strs = []
+    for face in room.faces:
+        room_strs.append(face_to_rad(face, blk, minimal))
+    for shd in room.shades:
+        room_strs.append(shade_to_rad(shd, blk, minimal))
+    return '\n\n'.join(room_strs)
+
+
+def model_to_rad(model, blk=False, minimal=False):
```
mostaphaRoudsari: Parts of the writer module will probably change after we talk about the folder structure.

tests/face_extend_test.py
```diff
@@ -1,19 +1,141 @@
...
+
+
+def test_default_modifiers():
+    """Test the auto-assigning of modifiers by face type and boundary condition."""
+    vertices_wall = [[0, 0, 0], [0, 10, 0], [0, 10, 3], [0, 0, 3]]
+    vertices_floor = [[0, 0, 0], [0, 10, 0], [10, 10, 0], [10, 0, 0]]
+    vertices_roof = [[10, 0, 3], [10, 10, 3], [0, 10, 3], [0, 0, 3]]
+
+    wf = Face.from_vertices('wall', vertices_wall)
+    assert wf.properties.radiance.modifier.name == 'generic_wall_0.50'
```
mostaphaRoudsari: A better test is probably comparing the full modifier which we can import from modifier set. In that case we don't have to update the name if we decide to change the modifier in the modifier-set at some point.

honeybee_radiance/modifierset.py
```diff
@@ -86,6 +89,9 @@ def __init__(self, name, wall_set=None, floor_set=None, roof_ceiling_set=None,
                 If None, it will be the honeybee generic default DoorSet.
             shade_set: An optional ShadeSet object for this ModifierSet.
                 If None, it will be the honeybee generic default ShadeSet.
+            air_wall_modifier: An optional Modifier to be used for all Faces with
```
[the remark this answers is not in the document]
chriswmackey: I'm just noting that we discussed this and it unfortunately seems that the term "AirWall" is so ingrained in the modeling culture that trying to change it to something like "AirFace" or "AirSurface" is going to result in more confusion than clarity. I'll note that DesignBuilder tried to coin the term "VirtualPartition" for this case but I worry that using the same terminology here might either be lost on users or it will imply that we are using DesignBuilders odd way of modeling these surfaces in E+ (adding windows to them and not using E+'s AirWallConstruction).

honeybee_radiance/modifierset.py
```diff
@@ -197,17 +219,21 @@ def modifiers(self):
...
     def modified_modifiers(self):
         """List of all modifiers that are not defaulted within the set."""
-        return self.wall_set.modified_modifiers + \
+        modified_modifiers = self.wall_set.modified_modifiers + \
             self.floor_set.modified_modifiers + \
             self.roof_ceiling_set.modified_modifiers + \
             self.aperture_set.modified_modifiers + \
             self.door_set.modified_modifiers + \
             self.shade_set.modified_modifiers
+        if self._air_wall_modifier is not None:
```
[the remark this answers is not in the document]
chriswmackey: Like all of the other modifiers on the ModifierSet, when there's no specific modifier assigned for air walls, it defaults to the global modifier, which is a 100% transmittance surface. This option is mostly here so that people can change the air wall modifier if they want.

honeybee_radiance/properties/aperture.py
```diff
@@ -0,0 +1,186 @@
...
+
+    __slots__ = ('_host', '_modifier', '_modifier_blk')
+
+    def __init__(self, host, modifier=None, modifier_blk=None):
+        """Initialize Aperture radiance properties.
+
+        Args:
+            host: A honeybee_core Aperture object that hosts these properties.
+            modifier: A Honeybee Radiance Modifier object for the aperture. If None,
+                it will be set by the parent Room ModifierSet or the the Honeybee
```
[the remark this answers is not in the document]
chriswmackey: Done.

honeybee_radiance/properties/face.py
chriswmackey: I wasn't sure whether we would need these classes to be more different once we implemented states. Now that I understand that a state is like a whole other copy of the object (e. Aperture, Shade) with its own modifiers, geometry, and assigned objects changed, I think we can have a base class here.

honeybee_radiance/writer.py
chriswmackey: So we decided in our talk that the current `to.rad` methods can remain the way that they are (dumping the whole object into a single Rad string). However, we will add a separate `Model.to.rad_folder` method to the writer that spits out the model in the folder structure.

tests/face_extend_test.py
chriswmackey: Fair point. I'll change it.

honeybee_radiance/properties/face.py
chriswmackey: Done!

PR #39, opened 2020-02-13
feat(lib): Load modifiers and modifier sets from files

chriswmackey:
This commit changes the libraries to load up from .rad, .mat, or .json files. The data follows a similar structure to honeybee_energy such that we can eventually combine these files with honeybee_energy into a single honeybee_standards repo.

It also adds a config file that allows users to change the folder from which the libraries load up.

### DIFF
--- a/honeybee_radiance/config.py
+++ b/honeybee_radiance/config.py
@@ -1,0 +1,309 @@
+"""Honeybee_radiance configurations.
+
+Import this into every module where access configurations are needed.
+
+Usage:
+
+.. code-block:: python
+
+    from honeybee_radiance.config import folders
+    print(folders.radiance_path)
+    folders.radiance_path = "C:/Radiance/bin"
+"""
+import honeybee.config as hb_config
+
+import os
+import platform
+import json
+
+
+class Folders(object):
+    """Honeybee_radiance folders.
+
+    Args:
+        config_file: The path to the config.json file from which folders are loaded.
+            If None, the config.json module included in this package will be used.
+            Default: None.
+        mute: If False, the paths to the various folders will be printed as they
+            are found. If True, no printing will occur upon initialization of this
+            class. Default: True.
+
+    Properties:
+        * radiance_path
+        * radbin_path
+        * radlib_path
+        * standards_data_folder
+        * modifier_lib
+        * modifierset_lib
+        * config_file
+        * mute
+    """
+
+    def __init__(self, config_file=None, mute=True):
+        # set the mute value
+        self.mute = bool(mute)
+
+        # load paths from the config JSON file 
+        self.config_file  = config_file
+
+    @property
+    def radiance_path(self):
+        """Get or set the path to Radiance installation folder.
+        
+        This is the top level folder that contains both the "bin" and the "lib"
+        directories.
+        """
+        return self._radiance_path
+    
+    @radiance_path.setter
+    def radiance_path(self, r_path):
+        exe_name = 'rad.exe' if os.name == 'nt' else 'rad'
+        if not r_path:  # check the PATH and then the default installation locations
+            r_path, rad_exe_file = self._which(exe_name)
+            if r_path is None:  # search within the default installation locations
+                r_path, rad_exe_file = self._find_radiance_folder()
+        else:
+            r_path = os.path.join(r_path, 'bin')
+            rad_exe_file = os.path.join(r_path, exe_name)
+
+        if r_path:  # check that the Radiance executable exists in the installation
+            assert os.path.isfile(rad_exe_file), \
+                '{} is not a valid path to a Radiance installation.'.format(r_path)
+
+        # set the radiance_path
+        self._radbin_path = r_path
+        if r_path:
+            self._radiance_path = os.path.split(r_path)[0]
+            self._radlib_path = os.path.join(self._radiance_path, 'lib')
+            if not self.mute:
+                print("Path to Radiance is set to: %s" % self._radiance_path)
+        else:
+            self._radiance_path = None
+            self._radlib_path = None
+
+    @property
+    def radbin_path(self):
+        """Get the path to Radiance bin folder.
+        
+        This is the "bin" directory for Radiance installation (the one that
+        contains the executable files).
+        """
+        return self._radbin_path
+
+    @property
+    def radlib_path(self):
+        """Get the path to Radiance lib folder."""
+        return self._radlib_path
+
+    @property
+    def standards_data_folder(self):
+        """Get or set the path to the folder standards loaded to honeybee_radiance.lib.
+        
+        This folder must have the following sub-folders in order to be valid:
+            * modifiers - folder with RAD files for modifiers.
+            * modifiersets - folder with JSON files of abridged ModifierSets.
+        """
+        return self._standards_data_folder
+    
+    @standards_data_folder.setter
+    def standards_data_folder(self, path):
+        if not path:  # check the default locations of the template library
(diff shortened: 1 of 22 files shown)


chriswmackey:
BTW, I will get the coverage up before we merge.

honeybee_radiance/config.py
```diff
@@ -0,0 +1,308 @@
...
+            for f in os.listdir('/Applications/'):
+                if f.lower() == 'radiance' and os.path.isdir('/Applications/{}'.format(f)):
+                    rad_path = '/Applications/{}'.format(f)
+                    break
+        elif platform.system() == 'Linux':  # search the usr/local folder
+            for f in os.listdir('/usr/local/'):
+                if f.lower() == 'radiance' and os.path.isdir('/usr/local/{}'.format(f)):
+                    rad_path = '/usr/local/{}'.format(f)
+                    break
+        else:  # then check the Radiance that comes with OpenStudio
```
mostaphaRoudsari: Should the check here be be `if rad_paths is None`? Right now you will probably never hit this condition.

honeybee_radiance/lib/modifiers.py
```diff
@@ -4,68 +4,187 @@
...
-generic_floor = Plastic.from_single_reflectance('generic_floor_0.20', 0.2)
-generic_wall = Plastic.from_single_reflectance('generic_wall_0.50', 0.5)
-generic_ceiling = Plastic.from_single_reflectance('generic_ceiling_0.80', 0.8)
-generic_door = Plastic.from_single_reflectance('generic_opaque_door_0.50', 0.5)
+try:
+    generic_floor = _rad_modifiers['generic_floor_0.20']
+except KeyError:
+    generic_floor = Plastic.from_single_reflectance('generic_floor_0.20', 0.2)
+    generic_floor.lock()
+    _rad_modifiers['generic_floor_0.20'] = generic_floor
```
mostaphaRoudsari: I'm a bit concerned about this approach to be a problem. The reason being there are two places that the default material might get assigned. What will be the case that default key is not available and we thing it should be considered ok? I think if the default key is not available we should throw an exception and ask for a fix.

honeybee_radiance/lib/_loadmodifiers.py
```diff
@@ -0,0 +1,39 @@
...
+                data = json.load(json_file)
+            for mod_name in data:
+                try:
+                    mod_dict = data[mod_name]
+                    m_class = modifier_class_from_type_string(mod_dict['type'].lower())
+                    mod = m_class.from_dict(mod_dict)
+                    mod.lock()
+                    _rad_modifiers[mod.name] = mod
+                except KeyError:
+                    pass  # not a Honeybee JSON file with modifiers
```
mostaphaRoudsari: Should we raise an error and say this object is not a modifier and must be removed?

honeybee_radiance/primitive.py
```diff
@@ -424,22 +424,22 @@ def duplicate(self):
     @staticmethod
     def filter_dict_input(input_dict):
         """Filter a dictionary of a Primitive to get modifier and dependency objects."""
-        try:  # ensure the putil module is imported, which imports all primitive modules
-            putil
+        try:
```
mostaphaRoudsari: A note here will be helpful to know why you try to access the module and then import it hasn't been imported already.

tests/assets/model/test_model.rad
```diff
@@ -4,65 +4,65 @@
...
 
-void glass generic_interior_window_vis_0.88
+void glass generic_exterior_window_vis_0.64
 0
 0
-3 0.9584154328610596 0.9584154328610596 0.9584154328610596
+3 0.6975761815384331 0.6975761815384331 0.6975761815384331
 
-void plastic generic_interior_shade_0.50
+void glass CustomTriplePane
```
mostaphaRoudsari: ?
```suggestion
void glass custom_triple_pane_0.3
```

tests/assets/model/test_model.rad
```diff
@@ -4,65 +4,65 @@
...
 
-void plastic generic_opaque_door_0.50
+void glass air_boundary
 0
 0
-5 0.5 0.5 0.5 0.0 0.0
+3 1.0 1.0 1.0
 
-void plastic IndoorLightShelf
+void plastic OutdoorLightShelf
```
mostaphaRoudsari:
```suggestion
void plastic outdoor_light_shelf_0.5
```

tests/assets/model/test_model.rad
```diff
@@ -4,65 +4,65 @@
...
+5 0.8 0.8 0.8 0.0 0.0
 
-void plastic generic_wall_0.50
+void plastic generic_interior_shade_0.50
 0
 0
 5 0.5 0.5 0.5 0.0 0.0
 
-void plastic OutdoorLightShelf
+void plastic IndoorLightShelf
```
mostaphaRoudsari:
```suggestion
void plastic indoor_light_shelf_0.70
```

honeybee_radiance/config.py
chriswmackey: Yes, I realized this error in my sleep last night. I'll fix it now.

honeybee_radiance/lib/modifiers.py
chriswmackey: For all cases where the modifiers aren't critical to the functioning of the rest of the application, I agree that your approach makes sense. But people won't be able to import `honeybee_radiance` if one of these modifiers is missing since ModifierSets won't have the default modifiers they need and, by extension, the FaceProperties won't be importable. So there aren't any cases where it should be considered ok to have one of these particular modifiers missing.

Also, I know that there will be several common cases where we won't have the necessary .rad files to load the modifiers if we want to separate them into their own `honeybee_standards` package that we can install in a central user library that doesn't get overwritten with each update.  I guess we could just duplicate these .rad files between the new `honeybee_standards` repo and `honeybee_radiance` so that there's always a backup copy to load from if we don't find the user library. So maybe the choice is between this type of rad-file duplication or the one we currently have here where we catch the missing modifiers and create them in Python if we don't find them. Just knowing how critical it is to have these modifiers, I would rather do the latter. What do you think?

honeybee_radiance/primitive.py
chriswmackey: Right. The original reason has changed since, before this PR, it was done because it would import all of the modifier classes and it wasn't clear that we always wanted that. Now that we have to import all of the modifier classes in order to correctly parse the user libraries, the reason why it's in a try/except is to avoid a circular reference. I'll add a note about it.

honeybee_radiance/lib/_loadmodifiers.py
chriswmackey: Ok. This type of exception could be helpful since we could provide additional information in the message about how to format the data correctly.

honeybee_radiance/lib/modifiers.py
mostaphaRoudsari: > Also, I know that there will be several common cases where we won't have the necessary .rad files to load the modifiers if we want to separate them into their own honeybee_standards package that we can install in a central user library that doesn't get overwritten with each update.

I'm not sure if I fully understand why this should cause such a case when we won't have the modifiers.

I think this needs a longer conversation to figure out the best solution. What you have implemented right now suffice for this PR. Let's revisit this after we talk if necessary.

tests/assets/model/test_model.rad
chriswmackey: This is just a test file but, ok. I guess the tests sent an example for others.

PR #41, opened 2020-02-20
Format docstrings for correct display in sphinx-generated bootstrap pages

santiagogaray:
Hi @chriswmackey 

Here's the first pass to the library.  There are some from_dict() functions that I will like to add examples but this is a good point to check things
.
Feel free to take a look at it and please make any comments about the changes, I did have to make a few decisions that were unique but nothing too out of the ordinary.


santiagogaray:
Hi @chriswmackey 
I took a stab at the missing dictionaries, please take a look at them when you get a chance.

chriswmackey:
Thanks @santiagogaray . I'll review this now.

honeybee_radiance/modifier/material/antimatter.py
```diff
@@ -9,9 +9,11 @@
 class Antimatter(Material):
     """Radiance Antimatter Material.
 
-    Antimatter is a material that can "subtract" volumes from other volumes. A ray
+    Antimatter is a material that can \'subtract\' volumes from other volumes. A ray
```
chriswmackey: This is a good thing to be aware of. Let's make sure that we document this in the wiki.

Also, I realized that this class lacks `Args` and `Properties`.

honeybee_radiance/modifier/material/ashik2.py
```diff
@@ -15,10 +15,37 @@ class Ashik2(Material):
...
+        name: Primitive name as a string. Cannot contain white spaces or special
+            characters.
+        modifier: Modifier. It can be primitive, mixture, texture or pattern.
+            (Default: "void").
+        values: An array 3 arrays for primitive data. Each of the 3 sub-arrays
+            refer to a line number in the radiance primitve definitions and the
+            values in each array correspond to values occurring within each line.
+            For example, [[], [], ['0.500', '0.500', '0.500', '0.000', '0.050']]
+            corresponds to values one would find for a Plastic material.
+            (Default: [[], [], []]).
```
chriswmackey: I agree that adding the `Args` to all of these modifier classes is a good idea but this example given here doesn't fit for these specific materials. So can we remove this text from all of the modifier docstrings that have it?
```
For example, [[], [], ['0.500', '0.500', '0.500', '0.000', '0.050']]
corresponds to values one would find for a Plastic material.
```

honeybee_radiance/modifier/material/metal2.py
```diff
@@ -11,7 +11,6 @@ class Metal2(Material):
 
     Metal2 is the same as plastic2, except that the highlights are modified by the
     material color.
-
```
chriswmackey: Is there a reason why you didn't implement `Args` and `Properties` for this class?

honeybee_radiance/modifier/material/mist.py
```diff
@@ -28,15 +28,15 @@ class Mist(Material):
     medium (no absorption). The scattering eccentricity parameter will likewise override
     the global setting if it is present. Scattering eccentricity indicates how much
     scattered light favors the forward direction, as fit by the Henyey-Greenstein
-    function:
+    function::
```
chriswmackey: Does the formula show up ok with just the `::`? If so, this would also be something good to document in the wiki.

Also I realized that this class also lacks `Args` and `Properties`.

honeybee_radiance/modifier/material/dielectric.py
```diff
@@ -17,11 +17,12 @@ class Dielectric(Material):
     changes as a function of wavelength. It is usually zero. (A pattern modifies only the
     refracted value.)
 
+    .. code-block:: shell
+
```
chriswmackey: This class lacks `Args` and `Properties`.

honeybee_radiance/modifier/material/illum.py
```diff
@@ -15,6 +15,8 @@ class Illum(Material):
     becomes invisible (if no string argument is given, or the argument is "void").
     Secondary sources are useful when modeling windows or brightly illuminated surfaces.
 
+    .. code-block:: shell
+
```
chriswmackey: This class lacks `Args` and `Properties`.

honeybee_radiance/modifier/material/interface.py
```diff
@@ -13,11 +13,12 @@ class Interface(Material):
     coefficient and refractive index are for the inside; the second ones are for the
     outside. Ordinary dielectrics are surrounded by a vacuum (1 1 1 1).
 
+    .. code-block:: shell
+
```
chriswmackey: This class lacks `Args` and `Properties`.

honeybee_radiance/modifier/material/plasdata.py
```diff
@@ -13,6 +13,8 @@ class Plasdata(Material):
     interpolated data. The arguments to this material are the data file and coordinate
     index functions, as well as a function to optionally modify the data values.
 
+    .. code-block:: shell
+
```
chriswmackey: This class lacks `Args` and `Properties`.

honeybee_radiance/modifier/material/plasfunc.py
```diff
@@ -14,6 +14,8 @@ class Plasfunc(Material):
     to this material include the color and specularity, as well as the function defining
     the specular distribution and the auxiliary file where it may be found.
 
+    .. code-block:: shell
+
```
chriswmackey: This class lacks `Args` and `Properties`.

honeybee_radiance/modifier/material/plastic2.py
```diff
@@ -14,19 +14,20 @@ class Plastic2(Material):
...
+    or functions are required), a period (\'.\') may be given in its place. (See the
     discussion of Function Files in the Auxiliary Files section). The urough value
     defines the roughness along the u vector given projected onto the surface. The vrough
     value defines the roughness perpendicular to this vector. Note that the highlight
     will be narrower in the direction of the smaller roughness value. Roughness values of
     zero are not allowed for efficiency reasons since the behavior would be the same as
     regular plastic in that case.
 
+    .. code-block:: shell
+
```
chriswmackey: This class lacks `Args` and `Properties`.

honeybee_radiance/modifier/material/prism1.py
```diff
@@ -17,6 +17,8 @@ class Prism1(Material):
     virtual light sources. The arguments give the coefficient for the redirected light
     and its direction.
 
+    .. code-block:: shell
```
chriswmackey: This class lacks `Args` and `Properties`.

honeybee_radiance/modifier/material/prism2.py
```diff
@@ -12,11 +12,12 @@ class Prism2(Material):
     The material prism2 is identical to prism1 except that it provides for two ray
     redirections rather than one.
 
+    .. code-block:: shell
```
chriswmackey: This class lacks `Args` and `Properties`.

honeybee_radiance/modifier/material/spotlight.py
```diff
@@ -14,11 +14,12 @@ class Spotlight(Material):
     are given. The length of the orientation vector is the distance of the effective
     focus behind the source center (i.e., the focal length).
 
+    .. code-block:: shell
```
chriswmackey: This class lacks `Args` and `Properties`.

honeybee_radiance/modifier/material/trans2.py
```diff
@@ -13,11 +13,12 @@ class Trans2(Material):
     plastic2, and the real arguments are the same as for trans but with an additional
     roughness value.
 
+    .. code-block:: shell
+
```
chriswmackey: This class lacks `Args` and `Properties`.

honeybee_radiance/modifier/material/transdata.py
```diff
@@ -12,13 +12,14 @@ class Transdata(Material):
     Transdata is like plasdata but the specification includes transmittance as well as
     reflectance. The parameters are as follows.
 
+    .. code-block:: shell
+
```
chriswmackey: This class lacks `Args` and `Properties`.

honeybee_radiance/modifier/material/transfunc.py
```diff
@@ -13,6 +13,8 @@ class Transfunc(Material):
     distribution as well as a reflectance distribution. Both reflectance and
     transmittance are specified with the same function.
 
+    .. code-block:: shell
+
```
chriswmackey: This class lacks `Args` and `Properties`.

honeybee_radiance/modifier/material/antimatter.py
santiagogaray: Indeed, I am making notes as I go through them, this one was sneaky!

santiagogaray: In regards to the classes without Arguments/Properties, I actually had Args/Properties in all of them and deleted them after. Bad judgement!
I had the feeling this classes weren't ready to be used, sort of a placeholder, and that those properties will probably change.  Clearly they may still be declared by users even as is.

santiagogaray: I guess I realize now I forgot to delete the Args/Properties from some classes, that's why you saw the inconsistency. Glad you found them useful!

honeybee_radiance/modifier/material/metal2.py
santiagogaray: Actually, these classes not implemented, in my jugement, weren't supposed to have Args/Properties, As I mentioned above, I just forgot to deleted them all!

honeybee_radiance/modifier/material/mist.py
santiagogaray: Yes, this is a variation of the `.. code-block:: shell` directive. I can document these options together.

santiagogaray:
Thanks @chriswmackey for the comments. All the materials should have the arguments and properties  now.I believe the sun path class wea argument is the only thing to correct. Should this be north instead?

honeybee_radiance/modifier/material/antimatter.py
chriswmackey: You are right on all counts. These classes are technically placeholders that should hopefully get fleshed out into something more detailed in the future. But they are still usable in their current state so it's helpful to document the Args and Properties for the time being.

chriswmackey:
Thanks for fixing everything @santiagogaray . And sorry I forgot about the sunpath question, You are right that this docstring is incorrect. It looks like it's a leftover from a previous iteration of the class If you can fix that, we should be good to merge.

santiagogaray:
Great @chriswmackey,  sunpath docstring fixed!

chriswmackey:
Thanks @santiagogaray . I'm going to merge this in now.

BTW, it would be very helpful if you can look over the docs of this repo:
https://github.com/ladybug-tools/ladybug-rhino
It's a small one and I recently generated the files from your updated ladybug-tools-template but it has dependencies on Rhino SDK and I know that this is causing the sphinx docs to not build correctly for some pages.  If you are able to find a solution to the issue within Sphinx, that would be ideal. Otherwise, I have some ideas for workarounds.

chriswmackey:
Side note about ladybug-rhino: the best workaround that I can think of if there's no way to have sphinx run without preserving the import statements is to put the imports of the Rhino SDK inside try/except statements. This should be quick to implement if any alternatives don't work out.

PR #48, opened 2020-04-07
fix(void): Change Void object return dict in to_dict method.

saeranv:
@chriswmackey look good?


saeranv:
@chriswmackey  I see the failing tests, I'll fix them and resend this.

saeranv:
This turned out to be a little more complicated then I initially thought. Issue posted here: https://github.com/ladybug-tools/honeybee-radiance/issues/49

saeranv:
@mostaphaRoudsari here are the changes we worked on. The `shade_extend_test.py` file is the specific failing test that we were trying to debug, but most tests are failing. 

Thanks!

PR #61, opened 2020-04-11
fix(cli-sunpath): remove analysis period options from sunpath wea subcommand

mostaphaRoudsari:
Since one can generate the wea file for part of the year that should be handled before using the honeybee-radiance command.

I realized the we have not really reviewed the CLI PRs. Starting from this one let's do the reviews and slowly develop a standard that we can enforce among all the projects.

### DIFF
--- a/honeybee_radiance/cli/sunpath.py
+++ b/honeybee_radiance/cli/sunpath.py
@@ -4,14 +4,13 @@ try:
     import click
 except ImportError:
     raise ImportError(
-        'click is not installed. Try `pip install . [cli]` command.'
+        'click is not installed. Try `pip install -U .[cli]` command.'
     )
 
 import sys
 from honeybee_radiance.lightsource.sunpath import Sunpath
 from ladybug.location import Location
 from ladybug.wea import Wea
-from ladybug.epw import EPW
 import logging
 import json
 import os
@@ -29,7 +28,7 @@ def sunpath():
 @sunpath.command('location')
 @click.option('--lat', default=0, type=float, show_default=True,
     help='The latitude of the location in degrees. Values must be between -90 and 90.'
-    ' Default is set to the equator.')
+         'Default is set to the equator.')
 @click.option('--lon', default=0, type=float, show_default=True,
     help='The longitude of the location in degrees')
 @click.option('--tz', default=0, type=float, show_default=True,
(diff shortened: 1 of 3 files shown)


honeybee_radiance/cli/sunpath.py
```diff
@@ -27,44 +26,55 @@ def sunpath():
...
     ' positive values are to the East of Greenwich and negative values are to the West.'
     )
-@click.option('--north', default=0, type=float, show_default=True,
+@click.option(
+    '--north', default=0, type=float, show_default=True,
     help='Angle to north (0-360). 90 is west and 270 is east'
     )
-@click.option('--start-date', default='JAN-01', show_default=True,
+@click.option(
+    '--start-date', default='JAN-01', show_default=True,
```
AntoineDao, line 48: Is it worth considering merging start-date and start-time into one input variable `start-date-time`? Dito for end-date.

honeybee_radiance/cli/sunpath.py
```diff
@@ -156,32 +158,41 @@ def sunpath_from_wea(wea, north, folder, name, log_file, start_date, start_time,
...
 @sunpath.command('epw')
 @click.argument('epw', type=click.Path(exists=True))
-@click.option('--north', default=0, type=float, show_default=True,
+@click.option(
+    '--north', default=0, type=float, show_default=True,
     help='Angle to north (0-360). 90 is west and 270 is east'
     )
-@click.option('--start-date', default='JAN-01', show_default=True,
+@click.option(
+    '--start-date', default='JAN-01', show_default=True,
```
AntoineDao, line 166: Dito previous comment on `start-date-time`

honeybee_radiance/cli/sunpath.py
mostaphaRoudsari: That can give a wrong impression to people. JAN-01-8:00 to JAN-15-20:00 will give an impression that every hour between Jan 1st at 8am to Jan 15 and 8pm will be included. However the time that will be included is the hours between 8am-8pm between Jan 1 to Jan 15. For instance Jan1 at 10 pm is not included.

mostaphaRoudsari:
> I personally prefer using flags for both mandatory and optional arguments as it removes any issues with argument ordering if we have more than one argument.

That's exactly one of those cases that I was hoping to come up from these reviews. I do put the mandatory ones as arguments and the optional ones as options. Mainly because that was the suggestion by Click manual and this is how Radiance command line works which I have been very happy with.

@chriswmackey, how have you been dealing with this subject?

honeybee_radiance/cli/sunpath.py
AntoineDao: Right, that makes sense!

PR #65, opened 2020-04-13
feat(material): add mirror material to modifiers

### LINKED ISSUE #46: Add Mirror Material?
I realize that there may be a reason for this but I know that I will need a mirror type of material to translate certain EnergyPlus materials into an equivalent Radiance materials. Are there any plans to add the [Radiance Mirror Material](https://floyd.lbl.gov/radiance/refer/ray.html#Mirror) to this library here?

mostaphaRoudsari:
resolves #46

### DIFF
--- a/honeybee_radiance/modifier/material/mirror.py
+++ b/honeybee_radiance/modifier/material/mirror.py
@@ -1,36 +1,45 @@
-"""Radiance Plastic Material.
+"""Radiance Mirror Material.
 
-http://radsite.lbl.gov/radiance/refer/ray.html#Plastic
+http://radsite.lbl.gov/radiance/refer/ray.html#Mirror
 """
 from __future__ import division
-
 from .materialbase import Material
 import honeybee.typing as typing
 
 
-class Plastic(Material):
-    """Radiance plastic material.
+class Mirror(Material):
+    """Radiance mirror material.
+
+    Mirror is used for planar surfaces that produce virtual source reflections. This
+    material should be used sparingly, as it may cause the light source calculation to
+    blow up if it is applied to many small surfaces. This material is only supported for
+    flat surfaces such as polygons and rings. The arguments are simply the
+    RGB reflectance values, which should be between 0 and 1.
+
+    An optional string argument may be used like the illum type to specify a different
+    material to be used for shading non-source rays. If this alternate material is given
+    as 'void', then the mirror surface will be invisible. This is only appropriate if
+    the surface hides other (more detailed) geometry with the same overall reflectance. 
 
     Args:
         identifier: Text string for a unique Material ID. Must not contain spaces
             or special characters. This will be used to identify the object across
             a model and in the exported Radiance files.
         r_reflectance: Reflectance for red. The value should be between 0 and 1
-            (Default: 0).
+            (Default: 1).
         g_reflectance: Reflectance for green. The value should be between 0 and 1
-            (Default: 0).
+            (Default: 1).
         b_reflectance: Reflectance for blue. The value should be between 0 and 1
-            (Default: 0).
-        specularity: Fraction of specularity. Specularity fractions greater than 0.1
-            are not realistic (Default: 0).
-        roughness: Roughness is specified as the rms slope of surface facets. A
-            value of 0 corresponds to a perfectly smooth surface, and a value of 1
-            would be a very rough surface. Roughness values greater than 0.2 are not
-            very realistic. (Default: 0).
+            (Default: 1).
         modifier: Material modifier (Default: "void").
+        alternate_material: An optional material may be used like the illum type to
+            specify a different material to be used for shading non-source rays. If this
+            alternate material is given as "void", then the mirror surface will be
+            invisible. This is only appropriate if the surface hides other (more
+            detailed) geometry with the same overall reflectance (Default: None).
         dependencies: A list of primitives that this primitive depends on. This
             argument is only useful for defining advanced primitives where the
-            primitive is defined based on other primitives. (Default: [])
+            primitive is defined based on other primitives (Default: []).
 
     Properties:
         * identifier
@@ -38,11 +47,10 @@ class Plastic(Material):
         * r_reflectance
         * g_reflectance
         * b_reflectance
-        * specularity
-        * roughness
         * average_reflectance
         * values
         * modifier
+        * alternate_material
         * dependencies
         * is_modifier
         * is_material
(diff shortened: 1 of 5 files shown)


chriswmackey:
Thanks for doing this, @mostaphaRoudsari . I will review now and merge. The conflict happened because I changed all of the cases where we used a default argument of `modifier="void"` to be `modifier=None`. Since we no longer expect people to put in strings for the `modifier` argument (it should either be None or it should be a true modifier instance like a `Void` instance).

honeybee_radiance/modifier/material/mirror.py
```diff
@@ -0,0 +1,332 @@
...
+        g_reflectance: Reflectance for green. The value should be between 0 and 1
+            (Default: 1).
+        b_reflectance: Reflectance for blue. The value should be between 0 and 1
+            (Default: 1).
+        modifier: Material modifier (Default: "void").
+        alternate_material: An optional material may be used like the illum type to
+            specify a different material to be used for shading non-source rays. If this
+            alternate material is given as "void", then the mirror surface will be
+            invisible. This is only appropriate if the surface hides other (more
+            detailed) geometry with the same overall reflectance (Default: None).
```
chriswmackey, line 39: It might be worth clarifying here that `"void"` is the same as `None`, which I think it is but I'm not sure. I'm also really unsure here if typical practice is to always put a material in here with the same rgb reflectance as the mirror? Or is it more common to leave this as None?

honeybee_radiance/modifier/material/mirror.py
```diff
@@ -0,0 +1,332 @@
...
+        mirror_material = Mirror("mirror_material", 0.95, .95, .95)
+        print(mirror_material)
+    """
+
+    __slots__ = ('_r_reflectance', '_g_reflectance', '_b_reflectance',
+                 '_alternate_material',)
+
+    def __init__(
+        self, identifier, r_reflectance=1.0, g_reflectance=1.0, b_reflectance=1.0,
+            modifier='void', alternate_material=None, dependencies=None):
```
chriswmackey, line 71: This should be `modifier=None` now.

honeybee_radiance/modifier/material/mirror.py
```diff
@@ -0,0 +1,332 @@
...
+    def b_reflectance(self, reflectance):
+        self._b_reflectance = \
+            typing.float_in_range(reflectance, 0, 1, 'blue reflectance')
+
+    @property
+    def alternate_material(self):
+        """Reflectance for red channel.
+
+        The value should be between 0 and 1 (Default: 0).
+        """
```
chriswmackey, line 134: Incorrect docstring here.

honeybee_radiance/modifier/material/mirror.py
```diff
@@ -0,0 +1,332 @@
...
+        "update value dictionaries."
+        if self.alternate_material is not None:
+            self._values[0] = [self.alternate_material.identifier]
+        self._values[2] = [
+            self.r_reflectance, self.g_reflectance, self.b_reflectance,
+        ]
+
+    @property
+    def r_reflectance(self):
+        """Reflectance for red channel.
```
chriswmackey, line 92: It would be helpful to write these doctrings as:
```
Get or set reflectance for red channel.
```
This way, users will be able to tell from the online docs that this property is set-able.

honeybee_radiance/modifier/material/mirror.py
```diff
@@ -0,0 +1,332 @@
...
+        """Get list of dependencies for this primitive.
+
+        Additional dependencies can be added with the add_dependent method.
+        """
+        return self._dependencies if not self.alternate_material \
+            else self._dependencies + [self.alternate_material]
+
+    @classmethod
+    def from_single_reflectance(
+        cls, identifier, rgb_reflectance=0.0, modifier='void', alternate_material=None,
```
chriswmackey, line 157: This should be `modifier=None`

honeybee_radiance/modifier/material/mirror.py
```diff
@@ -0,0 +1,332 @@
...
+            dependencies=None):
+        """Create mirror material with single reflectance value.
+
+        Args:
+            identifier: Text string for a unique Material ID. Must not contain spaces
+                or special characters. This will be used to identify the object across
+                a model and in the exported Radiance files.
+            rgb_reflectance: Reflectance for red, green and blue. The value should be
+                between 0 and 1 (Default: 0).
+            modifier: Material modifier (Default: "void").
```
chriswmackey, line 167: `(Default: None)`

honeybee_radiance/modifier/material/mirror.py
```diff
@@ -0,0 +1,332 @@
...
+    def from_primitive_dict(cls, primitive_dict):
+        """Initialize mirror from a primitive dict.
+
+        Args:
+            data: A dictionary in the format below.
+
+        .. code-block:: python
+
+            {
+            "modifier": "",  # primitive modifier (Default: "void")
```
chriswmackey, line 198: This should be:
```
"modifier": {},  # primitive modifier (Default: None)
```

honeybee_radiance/modifier/material/mirror.py
```diff
@@ -0,0 +1,332 @@
...
+        .. code-block:: python
+
+            {
+            "type": "mirror",  # Material type
+            "identifier": "",  # Material identifier
+            "display_name": string  # Material display name
+            "r_reflectance": float,  # Reflectance for red
+            "g_reflectance": float,  # Reflectance for green
+            "b_reflectance": float,  # Reflectance for blue
+            "modifier": {} or void,  # Material modifier
```
chriswmackey, line 271: This should be:
```
"modifier": {},  # Material modifier (Default: None)
```

chriswmackey:
I'm just going to close this PR as I merged the changed into [this one](https://github.com/ladybug-tools/honeybee-radiance/pull/68) and I addressed all of my comments.

honeybee_radiance/modifier/material/mirror.py
mostaphaRoudsari: As discussed it is not the same.

honeybee_radiance/modifier/material/mirror.py
mostaphaRoudsari: Thanks. Fixed.

mostaphaRoudsari:
I reverted back the other commits that was introducing bugs and will fix the comments that you made here and we can merge it afterwards.

PR #66, opened 2020-04-14
feat(model): Add method for Model.to.rad_folder

chriswmackey:
This commit adds a method to the radiance writer to serialize a honeybee model into a radiance folder structure. At the moment, there are no states in honeybee-radiance and so this method only serializes objects into the static and bsdf folders.

There are a number of things to note about this implementation:
* All aperture-assigned static shades are written into either the opaque or nonopaque folder (not the aperture folder) since these are the only folders that allow us to distinguish between indoor and outdoor shades (the static aperture folder does not support this distinction).
* We currently do not enforce that apertures need to have a nonopaque modifier. So opaque apertures and doors are still written into the aperture folder.
* For every object where there is an overridden modifier_blk, two completely new modifiers are generated for every unique combination of modifer and modifier_blk. This greatly increases the complexity of the code since this type of check has to be run for every single sub-directory in the radiance folder structure.  But it's the only way to support the flexibility pf overriding the modifier_blk while also getting the matching names between modifer and modifer_blk that the radiance folder structure expects.
* In the event that no objects of a certain type are in the model (like no nonopaque building faces), I currently don't write any files at all rather than writing empty files. This is likely to be a very common occurrence since the folder structure has many directories that will rarely get use.
* I have included checks to be sure that, for every Face, Aperture, and Door with a Surface boundary condition, only one of the pair of the matching objects is written. There's no logic to which of the two objects is picked beyond whichever shows up first in the list (so the direction that this interior object should be considered random ).

I still have to add some tests for the new method but the code here should be good enough to review, @mostapharoudsari.

### DIFF
--- a/honeybee_radiance/properties/model.py
+++ b/honeybee_radiance/properties/model.py
@@ -1,6 +1,7 @@
 # coding=utf-8
 """Model Radiance Properties."""
 from honeybee.extensionutil import model_extension_dicts
+from honeybee.boundarycondition import Surface
 
 from ..lib.modifiersets import generic_modifier_set_visible
 from ..lib.modifiers import black, generic_context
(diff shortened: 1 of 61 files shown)


mostaphaRoudsari:
Hey @chriswmackey,

I'm starting for your comments here:

> For every object where there is an overridden modifier_blk, two completely new modifiers are generated for every unique combination of modifer and modifier_blk. This greatly increases the complexity of the code since this type of check has to be run for every single sub-directory in the radiance folder structure. But it's the only way to support the flexibility pf overriding the modifier_blk while also getting the matching names between modifer and modifer_blk that the radiance folder structure expects.

Can you give me an example for this? There might be ways to avoid this by reusing the same material.

> I have included checks to be sure that, for every Face, Aperture, and Door with a Surface boundary condition, only one of the pair of the matching objects is written. There's no logic to which of the two objects is picked beyond whichever shows up first in the list (so the direction that this interior object should be considered random ).

This is fine until we get to the point that we need to model interior window groups. In that case, we have to write both or find a way to indicate which one in written.

honeybee_radiance/properties/model.py
```diff
@@ -304,11 +431,62 @@ def load_properties_from_dict(data):
...
+        child_shades = []
+        for room in host._rooms:
+            child_shades.extend(room._outdoor_shades)
+            for face in room.faces:
+                child_shades.extend(face._outdoor_shades)
+                for ap in face._apertures:
+                    child_shades.extend(ap._outdoor_shades)
+                for dr in face._doors:
+                    child_shades.extend(dr._outdoor_shades)
+        for face in host._orphaned_faces:
```
mostaphaRoudsari, line 478: Any specific reason that you use the private property when you get them from host? Shouldn't we just use the public method?

honeybee_radiance/properties/model.py
```diff
@@ -392,6 +570,29 @@ def _check_and_add_orphaned_shade_modifier(self, obj, modifiers):
             if not self._instance_in_array(generic_context, modifiers):
                 modifiers.append(generic_context)
 
+    @staticmethod
+    def _sort_shades(shades):
```
mostaphaRoudsari, line 574: It is semantics but to me this is more of a separate or get shades by opaque more then sort.

honeybee_radiance/writer.py
```diff
@@ -127,61 +137,356 @@ def room_to_rad(room, blk=False, minimal=False):
...
+    """
+    modifiers = []
+    for obj in geometry_objects:
+        mod = obj.properties.radiance.modifier
+        if not _instance_in_array(mod, modifiers):
+            modifiers.append(mod)
+    return list(set(modifiers))
+
+
+def _unique_modifier_blk_combinations(geometry_objects):
```
mostaphaRoudsari, line 410: Interesting! why do we need the combinations here? Can't we just look for unique blacked and normal modifiers?

chriswmackey:
To respond to your comments:

1. I can say that the current implementation of `modifier_blk` is as simple as it can be given the current radiance folder structure and the desire to be able to override modifier_blk on individual geometries. The only way that it could get simpler is by changing the Radiance folder structure so that the .blk modifiers didn't need to have the same identifier as the .mat modifiers or we get rid of the override capability. But I realize that either of these options create issues and so I understand that it is what it is. I just wanted to give you fair warning before you reviewed the code since dealing with modifier_blk means it takes more time to understand the code.

2. Ok.

honeybee_radiance/properties/model.py
```diff
@@ -161,6 +162,132 @@ def global_modifier_set(self):
...
+        """
+        opaque_faces = []
+        opaque_faces_blk = []
+        nonopaque_faces = []
+        nonopaque_faces_blk = []
+        interior_faces = set()
+        for face in self.host.faces:
+            if isinstance(face.boundary_condition, Surface):
+                if face.identifier in interior_faces:
+                    continue
```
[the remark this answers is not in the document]
chriswmackey: No, we are not assuming that adjacent surfaces will have the exact same identifier. That would be illegal in the honeybee schema for two objects to have the same identifier just as it was illegal for two objects to have the same name in the past.

If you look 2 lines after this one, you will see that we are adding the `face.boundary_condition.boundary_condition_object` to this `interior_faces` set. Not the identifier of the face itself. So the `interior_faces` set only tracks the boundary condition objects that have already been accounted for by their adjacent surface.

honeybee_radiance/properties/model.py
```diff
@@ -161,6 +162,132 @@ def global_modifier_set(self):
...
+                interior_faces.add(face.boundary_condition.boundary_condition_object)
+            if face.properties.radiance.is_opaque:
```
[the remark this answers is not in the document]
chriswmackey: I thought `continue` advances the `for` loop to the next cycle skipping over everything else in the current cycle. Let me test if I'm correct here.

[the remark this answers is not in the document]
chriswmackey: I double-checked and I am using `continue` correctly here. I think you are just misunderstanding what the code here is intended to do. It first checks that the face isn't already accounted for by it's neighbor if it has a Surface boundary condition. If it hasn't, then it goes and adds it to the appropriate opaque/non-opaque list.

honeybee_radiance/properties/model.py
chriswmackey: Sure. I did it to avoid the casting from the internal value (a list) to the public one (a tuple). But the extra time/memory it takes to do this casting once shouldn't make a difference. I'll change it back.

chriswmackey: Actually, I see that these hidden methods get called a lot and all of the casting to tuples is going to eat up time and memory. I will move them over into honeybee-core if these methods remain unchanged after we implement dynamic geometry. I was thinking that these methods might be temporary while I still don't understand if shades assigned to an aperture should be a part of the aperture or indoor/outdoor geometry.

honeybee_radiance/writer.py
chriswmackey: Here's the situation:
I have two opaque objects with the same `modifier` but, on one of them, I have overridden the `modifier_blk`. Technically, there's only one modifier that "needs" to be written into the .mat file but I can't do this because there are two different modifiers to be written into the .blk file (one completely black modifier and the, the other, the overridden `modifer_blk`). So, I can't write just one modifier in tho the .mat file because the .mat and .blk files need to contain the same number of modifiers and the names of the modifiers need to match between the .mat and the .blk file.

So I need to duplicate the original modifier, change it's identifier to match that of the overridden `modifier_blk` and write two "duplicate" modifiers into the .mat file. This type of duplication only needs to happen every time there's a unique combination of `modifier` and overridden `modifier_blk` (otherwise, I can just use the same `modifier` and `modifier_blk` pair that already exists).

I hope that makes it clear.

PR #68, opened 2020-04-17
fix(glass): Enforce use of float for refraction_index in dict

chriswmackey:
This commit also add the mirror Material for Mostapha's branch.


mostaphaRoudsari:
We should revert this PR.

As I mentioned in the other repository this change will make standard radiance material definitions invalid in honeybee. I see that you have updated the tests to pass with the change instead of thinking the test has a reason to be like that!

https://github.com/ladybug-tools/honeybee-radiance/pull/68/files#diff-807882db885f66a75f0ca68bd4ddcc16R118

PR #69, opened 2020-04-17
fix(mirror): Making Void an acceptable alternate material

chriswmackey:
I ran a test and found that the Mirror alaternate_material didn't properly allow Void to be assigned as an input to turn the material invisible. This commit fixes this issue.


mostaphaRoudsari:
We talked about this yesterday and this actually introduces a bug. The original code was the correct one.

Can I suggest to wait a day or two or ask for some input before just making major changes like this? In case it was really a bug it would have been good for me to know that there has been a bug. This is not good practice to just go ahead and merge changes to master because it was not what you assumed it should be. 😉

PR #70, opened 2020-04-18
feat(material): add mirror material to modifiers

### LINKED ISSUE #46: Add Mirror Material?

mostaphaRoudsari:
resolves #46

Also addressed all the comments from @chriswmackey in #65 and addressed the limitation that was pointed out in #69 but this iteration differs `void` from `None` for alternate material.

### DIFF
--- a/honeybee_radiance/modifier/material/__init__.py
+++ b/honeybee_radiance/modifier/material/__init__.py
@@ -14,6 +14,7 @@ from .metal import Metal
 from .metal2 import Metal2
 from .metdata import Metdata
 from .metfunc import Metfunc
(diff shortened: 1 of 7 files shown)


chriswmackey:
@mostaphaRoudsari ,

Ok. We will merge this and I will send a new PR with all of the other bug fixes that I sent after I merged in your old changes.

PR #71, opened 2020-04-18
docs(glass): Elaborate docstrings to explain implicit code

chriswmackey:
Accepting `None` for these cases is a bit implicit and it creates issues when parsing honeybee-schema since `None` indicates to a property that doesn't need to exist at all (rather than a number that needs to be there but just has a default value).

This commit also makes the docstrings of the Mirror alternate_material clearer.

### DIFF
--- a/honeybee_radiance/modifier/material/glass.py
+++ b/honeybee_radiance/modifier/material/glass.py
@@ -22,8 +22,9 @@ class Glass(Material):
             and 1 (Default: 0).
         b_transmissivity: Transmissivity for blue. The value should be between 0 and
             1 (Default: 0).
-        refraction: Index of refraction. 1.52 for glass and 1.4 for ETFE
-            (Default: 1.52).
+        refraction_index: Index of refraction. Typical values are 1.52 for float
+            glass and 1.4 for ETFE. If None, Radiance will default to using 1.52
+            for glass (Default: None).
         modifier: Material modifier (Default: None).
         dependencies: A list of primitives that this primitive depends on. This
             argument is only useful for defining advanced primitives where the
@@ -79,7 +80,7 @@ class Glass(Material):
 
     @property
     def r_transmissivity(self):
-        """Transmissivity for red channel.
+        """Get or set the transmissivity for red channel.
 
         The value should be between 0 and 1 (Default: 0).
         """
@@ -92,7 +93,7 @@ class Glass(Material):
 
     @property
     def g_transmissivity(self):
-        """Transmissivity for green channel.
+        """Get or set the transmissivity for green channel.
 
         The value should be between 0 and 1 (Default: 0).
         """
@@ -105,7 +106,7 @@ class Glass(Material):
 
     @property
     def b_transmissivity(self):
-        """Transmissivity for blue channel.
+        """get or set the transmissivity for blue channel.
 
         The value should be between 0 and 1 (Default: 0).
         """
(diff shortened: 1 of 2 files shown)


chriswmackey:
@mostaphaRoudsari ,

1. You clearly see that all of tests pass and that it is compatible with Radiance. You can parse a string without the 1.52 property just as easily as you can parse one that has it. The only change we are making here is that honeybee will explicitly write the 1.52 to radiance strings and dictionaries instead of writing `None` and having it be implicit. It's effectively treating refraction index the same way we treat the rgb transmittance. These rgb values may have a default of 0 but we still write them into the radiance string and dictionary if the user doesn't change the default. Do you not agree that explicit is better than implicit?

> None value for alternate material will keep the material as mirror 
2. That is what my doctring says but I'll change "will keep the same rgb relfectance" to "will keep the material as mirror" if that's the wording you are getting hung up on. We really need something here since you don't currently explain what `None` means in the docstring or explain that it is different from `void`.

mostaphaRoudsari:
> Do you not agree that explicit is better than implicit?

I agree with the statement but not the use case here. This is unnecessary information for most of the users. Just to ensure you that I understand what are you trying to do this is how the material was implemented in honeybee[+] radiance implementation originally: https://github.com/ladybug-tools/honeybee/blob/9c70b5d83df08e5ae2deff5cdb74b9e23e25c88b/honeybee/radiance/material/glass.py#L22-L36

I reverted that implementation to the current implementation  `None` is good and closely follows Radiance's approach to creating a glass material. I would leave it as is.

> That is what my doctring says

I'm not sure. See these two cases:

1. It is only correct for Void. None doesn't make it invisible.

  image

  https://github.com/ladybug-tools/honeybee-radiance/pull/71/files#diff-d1023f2a2f2bd648cdd8e650b1cfb2a0R22-R23

2. This is also incorrect. None doesn't make the mirror invisible.

  image

  https://github.com/ladybug-tools/honeybee-radiance/pull/71/files#diff-d1023f2a2f2bd648cdd8e650b1cfb2a0R135-R136

I saw that you have it right in the other parts of the file so they should be a typing mistake. I agree that the one here is an improvement to the current docstring.

https://github.com/ladybug-tools/honeybee-radiance/pull/71/files#diff-d1023f2a2f2bd648cdd8e650b1cfb2a0R37-R40

chriswmackey:
1. I reverted almost all of the changes I made to the tests in the hopes that this makes it clearer to you that the change isn't intended to change the radiance string parsing capability. The only thing that I am trying to change is make the default more explicit and write a number into the dictionary representation instead of `None`. I commented on the issue you reopened about why there needs to be a change to honeybee-radiance to support the Pydantic code that you put there.

2. Good catch. Sorry that I was only looking at the other docstring. It's fixed now.

chriswmackey:
1. I guess if you really feel strongly that `None` is what you want users to see whenever they check the `refraction_index` property on the Python object, I could just not write `refraction_index`  into the dictionary if it's `None` and then the Pydantic code can parse it. But we'll need to merge something in here if we want to do the Pydantic code correctly in a way that will also work on the honeybee-schema-dotnet side.

Let me know what you think about the changes I just made and we'll go from there.

mostaphaRoudsari:
1. Replied on the other issue. The code as is will work fine with Pydantic, OpenAPI and dotnet schema. Let's continue the conversation there.

2. Thank you for improving the docstrings. Agreed that the distinction between None and void is helpful.

chriswmackey:
Ok, this is not a docstring-only PR that just elaborates the docstring to explain what is going on in the implicit code.

I also fixed a number of docstrings that were pretty clearly copy/pasted and not updated such that they were incorrect.

PR #78, opened 2020-04-26
fix(bsdf): Ensure BSDF is serialize-able to JSON

chriswmackey:
This commit changes the BSDF dictionary to use a string to store the BSDF data instead of bytes, which are not serialize-able to JSON.

I was not able to compress the string further by removing white spaces since this messed with the re-serialization process (it could not identify the file type as TensorTree or Klems after all white space was removed). Maybe if we recorded the number of white spaces removed in each part so that we could put them back correctly, a strategy like that could work. But, at that point, I question whether we are actually saving much space in the resulting JSON..

### DIFF
--- a/honeybee_radiance/modifier/material/bsdf.py
+++ b/honeybee_radiance/modifier/material/bsdf.py
@@ -376,13 +376,12 @@ class BSDF(Material):
 
         fp = os.path.join(folder, '%s.xml' % data['identifier'])
         # write bytes to xml file
-        cls.decompress_bytes_to_file(data['bsdf_data'], fp)
+        cls.decompress_to_file(data['bsdf_data'], fp)
 
-        uo_dict = data['up_orientation']
         cls_ = cls(
             bsdf_file=fp,
             identifier=data['identifier'],
-            up_orientation=[uo_dict['x'], uo_dict['y'], uo_dict['z']],
+            up_orientation=data['up_orientation'],
             thickness=data['thickness'],
             modifier=modifier,
             dependencies=dependencies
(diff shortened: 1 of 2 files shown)


chriswmackey:
I also fixed the up-vector to be an array like all of our other vector representations across the schema.

tests/material_bsdf_test.py
```diff
@@ -61,3 +62,10 @@ def test_to_and_from_dict():
     with open(bsdf_in.bsdf_file, 'r') as f1, \
             open(bsdf_from_dict.bsdf_file, 'r') as f2:
         assert f1.read() == f2.read()
+
+
+def test_bsdf_to_json():
+    """Ensure that the BSDF dictionary is serialize-able to JSON."""
+    test_bsdf = BSDF(klems_bsdf_file)
+    bsdf_dict = test_bsdf.to_dict()
+    bsdf_str = json.dumps(bsdf_dict)
```
mostaphaRoudsari, line 71: Do you need to check something for the translation to ensure it has been successful?

honeybee_radiance/modifier/material/bsdf.py
```diff
@@ -376,13 +376,12 @@ def from_dict(cls, data, folder='.'):
 
         fp = os.path.join(folder, '%s.xml' % data['identifier'])
         # write bytes to xml file
-        cls.decompress_bytes_to_file(data['bsdf_data'], fp)
+        cls.decompress_to_file(data['bsdf_data'], fp)
```
mostaphaRoudsari, line 379: This will be an issue when we move the JSON file for the model around. Probably we should think of loading the data to the object and keeping it there until we write it to a folder / file.

tests/material_bsdf_test.py
chriswmackey: Sure. The `from_dict` test above this kinda already checks that part but I guess it doesn't hurt to do it again here.

honeybee_radiance/modifier/material/bsdf.py
chriswmackey: I think that is probably something that we could implement now that we have set honeybee up in a way that you can use the same instance of a BSDF to assign a modifier to a whole host of objects (so you don't end up duplicating the data in the XML for each object that the BSDF is assigned to). It's a lot to tackle in this PR, though. I will open an issue for it. https://github.com/ladybug-tools/honeybee-radiance/issues/80

mostaphaRoudsari: Sounds good! And I won't be worried about the duplication at this point if that is going to be too much of a hassle.

Issue #81, opened 2020-04-27
Change the Sensor Schema


chriswmackey:
@mostaphaRoudsari ,

I can see that the [Sensor Schema](https://github.com/ladybug-tools/honeybee-radiance/blob/master/honeybee_radiance/sensor.py#L81-L84) feels like an old relic from before we realized that we could use arrays instead of dictionaries to represent points and vectors. Should we change the schema now to just use two arrays of 3 floats in order to keep things consistent with the rest of our schemas? I feel that this will also make the sensor closer to how points and vectors for the views are documented.

mostaphaRoudsari:
Yes. Let's do this.

PR #83, opened 2020-04-27
fix(sensor): Tweak sensor schema to align with other objects

chriswmackey:
Resolves https://github.com/ladybug-tools/honeybee-radiance/issues/81

### DIFF
--- a/honeybee_radiance/sensor.py
+++ b/honeybee_radiance/sensor.py
@@ -7,21 +7,21 @@ class Sensor(object):
     """A radiance sensor.
 
     Args:
-        position: Position of sensor as (x, y, z) (Default: (0, 0, 0)).
-        direction: Direction of sensor as (x, y, z) (Default: (0, 0, 1)).
+        pos: Position of sensor as (x, y, z) (Default: (0, 0, 0)).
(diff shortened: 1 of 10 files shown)


honeybee_radiance/sensor.py
```diff
@@ -31,14 +31,13 @@ def from_dict(cls, sensor_dict):
         .. code-block:: python
 
             {
-            'x': float, 'y': float, 'z': float,
-            'dx': float, 'dx': float, 'dz': float
+            'position': [0, 0, 0],  # array of 3 numbers for the sensor position
+            'direction': [0, 0, 1]  # array of 3 numbers for the sensor direction
```
mostaphaRoudsari, line 35: Since there will be 1000s of sensors in most of studies I think keeping the keys shorter might be a good idea. We can use `pos` and `dir` if you think `x` and `dx` is very abstract.

chriswmackey: Ok. As long as `pos` and `dir` are also the names of the properties on the class, that is fine by me.

chriswmackey: I guess the one other thing to be aware of is that you are overwriting the native Python dir method when you make it a property on the class. But I guess we do this in other areas like for `face.type`.

PR #93, opened 2020-05-05
feat(stategeo): Add StateGeometry class to handle state-assigned shades

chriswmackey:
This commit creates a separate StateGeometry class for dynamic radiance shades and gets rid of the use of honeybee-core Shades for this purpose.

Resolves https://github.com/ladybug-tools/honeybee-schema/issues/112

Also, now that there are 3 different modules involved with dynamic geometry, I decided to put all of them in the same `dynamic` sub-package.

Lastly, made a small edit to the SensorGrid and View dictionaries: adding the 'type' key into the dictionary in order to keep it in line with all of the other schemas.

### DIFF
--- a/honeybee_radiance/dynamic/__init__.py
+++ b/honeybee_radiance/dynamic/__init__.py
@@ -1,0 +1,5 @@
+"""Objects for creating and exporting dynamic geometry."""
+
+from .group import DynamicShadeGroup, DynamicSubFaceGroup
+from .state import RadianceShadeState, RadianceSubFaceState
+from .stategeo import StateGeometry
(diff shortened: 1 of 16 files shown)


honeybee_radiance/dynamic/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Objects for creating and exporting dynamic geometry."""
+
+from .group import DynamicShadeGroup, DynamicSubFaceGroup
+from .state import RadianceShadeState, RadianceSubFaceState
```
mostaphaRoudsari, line 4: If they are Radiance only should we remove Radiance form in front of the names? Basically call the `ShadeState` and `SubFaceState`?

chriswmackey: I am a bit uncertain about whether we are going to create similar objects over on honeybee-energy, in which case we could have conflicting names of classes in the schema. It's possible that I end up settling on a different organization of objects over there but it seems safer to put `Radiance` in the name for now.

PR #101, opened 2020-05-12
feat(sensorgrid): Add SensorGrid and View room_idnetifier

chriswmackey:
This will help with minimizing simulation time by allowing us to export a JSON noting the aperture groups to be run with the grid.

I also added a method to the Room.radiance.properterites to generate a SensorGrid from the Room floors

Lastly, I deleted duplicate ToString methods that were on the SensorGrid and View classes.

### DIFF
--- a/honeybee_radiance/lightpath.py
+++ b/honeybee_radiance/lightpath.py
@@ -1,0 +1,86 @@
+"""Utilities to determine the path of light taken through interior spaces of a model."""
+from honeybee.boundarycondition import Surface
+
+
+def light_path_from_room(model, room_identifier, static_name='static_apertures'):
+    """Get the dynamic aperture groups that need to be simulated for a room in a model.
+    
+    Args:
+        model: A honeybee Model object which will be used to identify the aperture
+            groups that are needed to simulate a single room.
+        room_identifier: Text string for the identifer of the Room in the model
+            for which the light path will be computed.
+        static_name: An optional name to be used to refer to static apertures
+            found within the model. (Default: 'static_apertures').
+
+    Returns:
+        A list of lists where each sub-list contains the identifiers of the aperture
+        groups through which light is passing
+
+    Usage:
+
+    .. code-block:: python
+
+        from honeybee_radiance.lightpath import light_path_from_room
+        from honeybee.room import Room
+        from honeybee.model import Model
+        from ladybug_geometry.geometry3d.pointvector import Point3D
+
+        room1 = Room.from_box('Tiny_House_Room1', 5, 10, 3)
+        room1[2].apertures_by_ratio(0.4, 0.01)  # east interior window
+        room1[3].apertures_by_ratio(0.4, 0.01)  # outdoor south window
+        south_ap1 = room1[3].apertures[0]
+        south_ap1.properties.radiance.dynamic_group_identifier = 'SouthWindow1'
+
+        room2 = Room.from_box('Tiny_House_Room2', 5, 10, 3, origin=Point3D(5, 0, 0))
+        room2[4].apertures_by_ratio(0.4, 0.01)  # west interior window
+        room2[1].apertures_by_ratio(0.4, 0.01)  # outdoor north window
+        north_ap2 = room2[1].apertures[0]
+        north_ap2.properties.radiance.dynamic_group_identifier = 'NorthWindow2'
+
+        Room.solve_adjacency([room1, room2], 0.01)
+        model = Model('TinyHouse', [room1, room2])
+
+        print(light_path_from_room(model, room1.identifier))
+
+        >> [['SouthWindow1'], ['static_apertures', 'NorthWindow2']]
+    """
+    # get the Room object from the Model
+    room = model.get_rooms_by_identifier([room_identifier])[0]
+
+    # gather all of the dynamic groups that the Room has in its apertures
+    grp_ids = set()
+    adj_rooms = set()
+    for face in room.faces:
+        for s_face in face.apertures + face.doors:
+            if s_face.properties.radiance._dynamic_group_identifier:
+                grp_id = s_face.properties.radiance._dynamic_group_identifier
+            else:
+                grp_id = 'static_apertures'
+            if isinstance(s_face.boundary_condition, Surface):
+                adj_room = s_face.boundary_condition.boundary_condition_objects[-1]
+                adj_rooms.add((grp_id, adj_room))
+            else:
+                grp_ids.add(grp_id)
+
+    # if there are no interior apertures, return the list as it is
+    light_path = [[grp_id] for grp_id in grp_ids]
+    if len(adj_rooms) == 0:
+        return light_path
+
+    # loop through adjacent rooms and gather their aperture groups
+    # TODO: Make this part recursive to trace the light path more than one room away.
+    room_objs = model.get_rooms_by_identifier([room_tup[1] for room_tup in adj_rooms])
+    for room_obj, room_tup in zip(room_objs, adj_rooms):
+        rm_grp_ids = set()
+        for face in room_obj.faces:
+            for s_face in face.apertures + face.doors:
+                if s_face.properties.radiance._dynamic_group_identifier:
+                    rm_grp_ids.add(s_face.properties.radiance._dynamic_group_identifier)
+                else:
+                    rm_grp_ids.add('static_apertures')
+        base_grp_id = room_tup[0]
+        for g_id in rm_grp_ids:
+            if g_id != base_grp_id:  # group has already been accounted for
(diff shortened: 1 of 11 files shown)


honeybee_radiance/sensorgrid.py
```diff
@@ -190,6 +198,40 @@ def count(self):
...
+                the aperture groups that will be run with this sensor grid.
+        """
+        base = {'count': self.count}
+        if model:  # add the aperture_groups key
+            if self._room_identifier:  # add only the groups in the light path
+                base['aperture_groups'] = \
+                    light_path_from_room(model, self._room_identifier)
+            else:  # add all groups across the entire model
+                base['aperture_groups'] = \
+                    [[grp_id] for grp_id in model.subface_group_identifiers]
```
mostaphaRoudsari, line 232: This is probably not required. If the model is not provided the ray-tracing will happen for all the glass resources anyways.

honeybee_radiance/sensorgrid.py
```diff
@@ -190,6 +198,40 @@ def count(self):
...
+        number of aperture groups that have to be run with this sensor grid.
+        If None, the grid will be run with all aperture groups in the model.
+        """
+        return self._room_identifier
+
+    @room_identifier.setter
+    def room_identifier(self, n):
+        self._room_identifier = typing.valid_string(n)
+
+    def model_folder_dict(self, model=None):
```
mostaphaRoudsari, line 215: I would like some more generic name link `info_dict`. This information can be used in other cases more than just writing to a folder.

mostaphaRoudsari, line 215: I also think we should have an option to assign aperture group names directly.

The current method will work with the set-up of generating single sensor grid for a room perfectly but there can be other cases. Imagine an L shape room with two windows at the end. One wing will be mapped to one aperture group and the other wing will be a separate sensor-grid that will be mapped to the other one.

honeybee_radiance/sensorgrid.py
```diff
@@ -27,9 +30,10 @@ class SensorGrid(object):
         * sensors
         * positions
         * directions
+        * room_identifier
     """
 
-    __slots__ = ('_identifier', '_display_name', '_sensors')
+    __slots__ = ('_identifier', '_display_name', '_sensors', '_room_identifier')
```
mostaphaRoudsari, line 36: It should probably be an input for `aperture_group_identifiers`. We can have a method that figures out these groups based on a room_identifier and a model but as I discussed in another comment what really matters is the `aperture_group_identifiers`.

honeybee_radiance/properties/room.py
```diff
@@ -51,6 +52,50 @@ def modifier_set(self, value):
...
+        The output will also include a Mesh3D object with faces that align with
+        the positions of the SensorGrid.
+
+        Note that the x_dim and y_dim refer to dimensions within the XY coordinate
+        system of the floor faces's planes. So rotating the planes of the floor faces
+        will result in rotated grid cells.
+
+        Args:
+            x_dim: The x dimension of the grid cells as a number.
+            y_dim: The y dimension of the grid cells as a number. Default is None,
```
mostaphaRoudsari, line 67: very nice!

honeybee_radiance/lightpath.py
```diff
@@ -0,0 +1,83 @@
+"""Utilities to determine the path of light taken through interior spaces of a model."""
+from honeybee.boundarycondition import Surface
+
+
+def light_path_from_room(model, room_identifier):
```
mostaphaRoudsari, line 5: This is a great functionality and is only available thanks to the well structured model. Really like it. I can imagine similar methods can be useful for automatically setting up airflow networks or visualize the path for air.

One comment about fixing the name for `static_apertures`. Do you think it will be possible to change the default name here. If we want to keep it fixed which is fine to me then I should not include this in the folder `cfg` file which means user can never change it.

honeybee_radiance/sensorgrid.py
chriswmackey: Ok. I will cut it out, then. This should save some time in the export process.

honeybee_radiance/sensorgrid.py
chriswmackey: Sounds good. `info_dict` it is.

chriswmackey: Ok, I'll add an input for it but, given the only thing that this method does right now is auto-calculate the light path and write the number of points, adding an option to specify the light path is kinda a case where the input and output are really similar. I guess if we start adding more information into this dictionary, this will make more sense.

mostaphaRoudsari: I think this is part of the longer conversation of if we keep the `room_identifier` or the `aperture_group_identifiers`. See below.

I reviewed the code form bottom to top and the comments are reversed! You should probably start from the last comment and make it to the top. Sorry!

honeybee_radiance/sensorgrid.py
chriswmackey: I understand the desire to do that but it raises a number of workflow issues, which will ultimately make it very difficult to get this "light path" information in models exported from Grasshopper. Probably the biggest issue it creates is that you will only be able to generate grids that have correct "light path" information after you have a model where you have assigned all apertures and solved all adjacencies. If I generate the grid from a Room before I have added apertures, this "light path" information would basically be empty. Or it could be wrong if I have set apertures but not solved adjacencies. And there's no easy way to figure it out at a later point unless we test that a point is inside a Room volume (which we could do but it seems like this could be unreliable if custom-generated grids are also in the mix).

But, if I store the Room identifier instead of the light path information (like I have done here), I'll be able to figure out the light path correctly upon export, regardless of whether the Room had all of this apertures or adjacencies at the time I created the grid and assigned it to the Room.

I guess if you were ok holding off on the creation/assignment of grids until after you have a complete Model , your proposal could work. But this is probably going to limit the flexibility of working with grids in Grasshopper. So do you agree that the current setup is better at least for now?

honeybee_radiance/properties/room.py
chriswmackey: Thank you!

honeybee_radiance/lightpath.py
chriswmackey: Thank you!  Yes, there will be similar methods for doing this with Airflow Networks and for the thermal comfort maps, where we usually want to export models for view factor calculation in groups of rooms that are inter-connected by air walls.  At a certain point, I should probably write this method to be recursive (right now it only calculates the path up to one Room away). I'll add a `TODO` for this in the code.

Hm. Having `static_apertures` as a fixed name would probably make our lives easier but I don't know what the users will think about it. I can expose it as an option on this method for now and I trust your judgement about what to do with the `cfg` file.

honeybee_radiance/sensorgrid.py
chriswmackey: Done.

honeybee_radiance/sensorgrid.py
chriswmackey: I guess another option is to support both a `aperture_groups` property and a `room_identifier` property on the SensorGrid. And the hard-specified light path with the `aperture_groups` will override the auto-calculated path from the Room and Model on export. This is a lot like what you suggested in the comment above except that the `aperture_groups` are a property of the grid instead of an option on the info_dict method.

Would you prefer I implement something like that?

mostaphaRoudsari: Yes. That is what I was suggesting except. But in my scenario the `aperture_groups` are not really the full light-path. I think your suggestion to have both is a great starting point and then we will see how it will play out in real cases. I think in most cases that one overwrites the `aperture_groups` they know enough to provide the full light-path.

And while we are here should we call the input and the key in the dictionary `light_path` to be consistent?

mostaphaRoudsari: Just a reminder that whatever we do here should also be done for the view object.

chriswmackey: I was thinking that `light_path` made more sense as a name, given that it's much more than just `aperture_groups`.

So am I correct in understanding that we should add a `light_path` property to the `SensorGrid` and `View` classes?  And, if this `light_path` is not `None`, the light path is autocalculated using the `room_identifier`. And, if they are both `None`, there's no `light_path` key in the output JSON next to the .pts file so the grid will be run with all aperture groups in the model.

chriswmackey: I will just assume that my understanding is correct and implement it now.

chriswmackey: It has been implemented. I'll send a parallel PR over to honeybee-schema.

PR #116, opened 2020-05-17
fix(model): Update Model.to.rad_folder to work with recent edits

chriswmackey:
This commit brings honeybee-radiance in line with the recent refactor of honeybee-radiance folder.

I don't expect the edits to result in any significant performance improvements since roughly the same number of checks are still performed (there may be a small improvement form just writing fewer files, albeit with the same overall file contents).

However, the changes greatly simplified the code. Previously, the Model.to.rad_folder method was over 200 lines long, which was realistically as short as it could be given that there were 16 folders that had to be written to.

For the static files (where much of the new simplification occurred), I mostly stuck to writing three files into the folders (.rad, .mat, and .blk). However, I separated the scene in two: 3 envelope files (containing honeybee Faces) and 3 shade files (containing honeybee Shades). I did this since the to.rad_folder code was simpler this way and there may be some cases where this separation is useful to people looking over the file contents.

### DIFF
--- a/honeybee_radiance/writer.py
+++ b/honeybee_radiance/writer.py
@@ -1,9 +1,10 @@
 # coding=utf-8
 """Methods to write to rad."""
-from ladybug.futil import write_to_file_by_name, preparedir, nukedir
+from ladybug.futil import write_to_file_by_name, preparedir
 from honeybee.config import folders
 from honeybee.boundarycondition import Surface
 from honeybee_radiance_folder.folder import ModelFolder
+import honeybee_radiance_folder.config as folder_config
 
 from .geometry import Polygon
 from .modifier.material import BSDF
@@ -225,7 +226,7 @@ def model_to_rad(model, blk=False, minimal=False):
     return '\n\n'.join(model_str), '\n\n'.join(modifier_str)
 
 
-def model_to_rad_folder(model, folder=None, minimal=False):
+def model_to_rad_folder(model, folder=None, folder_type=2, minimal=False):
     r"""Write a honeybee model to a rad folder.
 
     The rad files in the resulting folders will include all geometry
@@ -238,6 +239,11 @@ def model_to_rad_folder(model, folder=None, minimal=False):
             Radiance folder. If None, the files will be written into a sub-directory
             of the honeybee-core default_simulation_folder. This sub-directory
             is specifically: default_simulation_folder/[MODEL IDENTIFIER]/Radiance
+        folder_type: An integer between 0-2 to note the type of folders to be
+            written out with the Model. (Default: 2). Choose from the following.
+                * 0: grid-based
+                * 1: view-based
+                * 2: includes both views and grids
         minimal: Boolean to note whether the radiance strings should be written
             in a minimal format (with spaces instead of line breaks). Default: False.
     """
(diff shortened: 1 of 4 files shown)


honeybee_radiance/writer.py
```diff
@@ -238,141 +239,79 @@ def model_to_rad_folder(model, folder=None, minimal=False):
...
     # write dynamic sub-face groups (apertures and doors)
     ext_dict = {}
+    out_subfolder = model_folder.aperture_group_folder(full=True)
     for group in model.properties.radiance.dynamic_subface_groups:
         if group.is_indoor:
             # TODO: Implement dynamic interior apertures once the radiance folder
             # structure is clear about how the "light path" should be input
             raise NotImplementedError('Dynamic interior apertures are not currently'
                                       ' supported by Model.to.rad_folder.')
         else:
```
mostaphaRoudsari, line 289: We can now support the indoor aperture group. You can just copy them to aperture_group_folder(full=True, indoor=True). With the addition of light-path to grids now we know how to treat them for simulations.

As for the direction that I know you asked about before I think we should have two and name them based on the rooms/faces that share this interior aperture. It will always be facing inwards towards that room similar to how it is currently written for the exterior aperture group.

honeybee_radiance/writer.py
```diff
@@ -238,141 +239,79 @@ def model_to_rad_folder(model, folder=None, minimal=False):
...
+    s_mods, s_mods_blk, mod_combs, mod_names = _collect_modifiers(shades, shades_blk)
+    _write_static_files(
+        folder, model_folder.scene_folder(full=True), '{}_shades'.format(model_id),
+        shades, shades_blk, s_mods, s_mods_blk, mod_combs, mod_names, False, minimal)
 
     # write dynamic sub-face groups (apertures and doors)
     ext_dict = {}
+    out_subfolder = model_folder.aperture_group_folder(full=True)
     for group in model.properties.radiance.dynamic_subface_groups:
         if group.is_indoor:
```
mostaphaRoudsari, line 284: Should I change the folder to also use `indoor` instead of interior? Right now we are using `indoor` for `dynamic scene` and `interior` for apertures.

honeybee_radiance/writer.py
chriswmackey: Ah, I didn't realize that. So the fact that we have the `light_path` that with the grid files means that we don't need to include this information within the interior apertures folder?  If so, that helps a lot.

A sample of what you mean by "both directions" would be really helpful. Does this mean that both directions go within the same .rad file? Or if they are separate, how are they differentiated from one another?

I agree that it's better to merge this now and, if you can add a sample or at least a better description to [this part of the radiance-folder-structure](https://github.com/ladybug-tools/radiance-folder-structure/tree/master/radiance_folder/model/aperture_group/interior), I will implement what you document.

honeybee_radiance/writer.py
chriswmackey: I think the difference in naming convention is ok because, for an aperture, we are talking about a part of the room envelope, which can be wither "exterior" envelope or "interior" envelope. For indoor shades, we are talking about things that are wholly inside the room envelope and not "interior" walls, which still go outside the "indoor" folder since they are a part of the room envelope.

Issue #88, opened 2020-05-04
Add CLI commands for translating Model to rad_folder and to rad


chriswmackey:
Now that the code exists to translate a honeybee model to the radiance folder structure, we should add commands [similar to the `translate` group in honeybee-energy](https://github.com/ladybug-tools/honeybee-energy/blob/master/honeybee_energy/cli/translate.py). So someone can pass in a model and get the radiance folder structure with the model in it. Or get a Radiance string.

PR #120, opened 2020-05-19
feat(cli): Add commands to translate model JSON to rad and rad_folder

chriswmackey:
This commit adds commands to translate Model JSONs to ordinance strings and radiance folders. These will be important for the testing the Rhino radiance plugin that @mingbopeng is working on.

I also fixes a few bugs in model re-serialization as part of this commit.

Resolves https://github.com/ladybug-tools/honeybee-radiance/issues/88

### DIFF
--- a/honeybee_radiance/cli/translate.py
+++ b/honeybee_radiance/cli/translate.py
@@ -1,0 +1,197 @@
+"""honeybee radiance translation commands."""
+
+try:
+    import click
+except ImportError:
+    raise ImportError(
+        'click is not installed. Try `pip install . [cli]` command.'
+    )
+
+from honeybee_radiance.reader import string_to_dicts
+from honeybee_radiance.mutil import dict_to_modifier, modifier_class_from_type_string
+
+from ladybug.futil import write_to_file_by_name
+from honeybee.model import Model
+
+import sys
+import os
+import logging
+import json
+
+_logger = logging.getLogger(__name__)
+
+
+@click.group(help='Commands for translating Honeybee JSON files to/from RAD.')
+def translate():
+    pass
+
+@translate.command('model-to-rad-folder')
+@click.argument('model-json')
+@click.option('--folder', help='Folder on this computer, into which the Radiance '
+              'folders will be written. If None, the files will be output in the'
+              'same location as the model_json.', default=None, show_default=True)
+@click.option('--folder-type', help='An integer between 0-2 to note the type of '
+              'folders to be written out with the Model.\n0: grid-based\n1: view-based'
+              '\n2: includes both views and grids', default=2, show_default=True)
+@click.option('--minimal', help='Boolean to note whether the radiance strings should '
+              'be written in a minimal format (with spaces instead of line breaks).',
+              default=False, show_default=True)
+@click.option('--log-file', help='Optional log file to output the progress of the'
+              'translation. By default this will be printed out to stdout',
+              type=click.File('w'), default='-')
+def model_to_rad_folder(model_json, folder, folder_type, minimal, log_file):
+    """Translate a Model JSON file into a Radiance Folder.
+    \n
+    Args:
+        model_json: Full path to a Model JSON file.
+    """
+    try:
+        # check that the model JSON is there
+        assert os.path.isfile(model_json), \
+            'No Model JSON file found at {}.'.format(model_json)
+
+        # set the default folder if it's not specified
+        if folder is None:
+            folder = os.path.split(os.path.abspath(model_json))[0]
+
+        # re-serialize the Model to Python
+        log_file.write('Re-serializing Model JSON.\n')
+        with open(model_json) as json_file:
+            data = json.load(json_file)
+        model = Model.from_dict(data)
+        log_file.write('Model re-serialization successful.\n')
+
+        # translate the model to a radiance folder
+        log_file.write('Translating Model to Radiance folder.\n')
+        rad_fold = model.to.rad_folder(model, folder, folder_type, minimal)
+        log_file.write('Model translation successful.\n')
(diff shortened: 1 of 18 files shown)


chriswmackey:
Also, I hope you don't mind, @mostaphaRoudsari , that I updated the two point in time sky classes to accept altitude values that are below the horizon. I remember that I used this in honeybee legacy to mode some twilight conditions.

honeybee_radiance/cli/translate.py
```diff
@@ -0,0 +1,197 @@
+"""honeybee radiance translation commands."""
+
+try:
+    import click
+except ImportError:
+    raise ImportError(
+        'click is not installed. Try `pip install . [cli]` command.'
```
mostaphaRoudsari, line 7: I think in most cases they actually need to download it from PyPI.

```suggestion
        'click is not installed. Try `pip install honeybee-radiance[cli]` command.'
```

honeybee_radiance/cli/translate.py
```diff
@@ -0,0 +1,197 @@
...
+@click.option('--folder-type', help='An integer between 0-2 to note the type of '
+              'folders to be written out with the Model.\n0: grid-based\n1: view-based'
+              '\n2: includes both views and grids', default=2, show_default=True)
+@click.option('--minimal', help='Boolean to note whether the radiance strings should '
+              'be written in a minimal format (with spaces instead of line breaks).',
+              default=False, show_default=True)
+@click.option('--log-file', help='Optional log file to output the progress of the'
+              'translation. By default this will be printed out to stdout',
+              type=click.File('w'), default='-')
+def model_to_rad_folder(model_json, folder, folder_type, minimal, log_file):
```
mostaphaRoudsari, line 42: Should we also add config file as an optional input?

honeybee_radiance/cli/translate.py
```diff
@@ -0,0 +1,197 @@
...
+        model_json: Full path to a Model JSON file.
+    """
+    try:
+        # check that the model JSON is there
+        assert os.path.isfile(model_json), \
+            'No Model JSON file found at {}.'.format(model_json)
+
+        # set the default folder if it's not specified
+        if folder is None:
+            folder = os.path.split(os.path.abspath(model_json))[0]
```
mostaphaRoudsari, line 55: What about using `dirname`?

```suggestion
            folder = os.path.dirname(model_json)
```

honeybee_radiance/cli/translate.py
```diff
@@ -0,0 +1,197 @@
...
+        with open(model_json) as json_file:
+            data = json.load(json_file)
+        model = Model.from_dict(data)
+
+        # translate the model to a rad string and put it in a dictionary
+        model_str, modifier_str = model.to.rad(model, blk, minimal)
+        rad_dict = {'model_str': model_str, 'modifier_str': modifier_str}
+
+        # write out the rad string
+        log_file.write(json.dumps(rad_dict))
```
mostaphaRoudsari, line 114: How do you imagine this JSON file to be used? Why not a standard rad file?

honeybee_radiance/cli/translate.py
```diff
@@ -0,0 +1,197 @@
...
+        sys.exit(1)
+    else:
+        sys.exit(0)
+
+@translate.command('modifiers-to-rad')
+@click.argument('modifier-json')
+@click.option('--minimal', help='Boolean to note whether the radiance strings should '
+              'be written in a minimal format (with spaces instead of line breaks).',
+              default=False, show_default=True)
+@click.option('--log-file', help='Optional RAD file to output the RAD string of the '
```
mostaphaRoudsari, line 126: I think naming them log-file can be confusing. They are really output files. I have used log-file for a command like separate sensor grid because the pts files is the real output and log-file logs the info about the grids. This is a generic comment about all the commands.

honeybee_radiance/cli/translate.py
chriswmackey: Hm. This would need to be exposed as an option on the writer.model_to_rad_folder method first. I guess I can do this as part of this PR.

honeybee_radiance/cli/translate.py
chriswmackey: `dirname` sounds good over `slpit` but I still need the `abspath` for most situations.

mostaphaRoudsari: I see. that works.

honeybee_radiance/cli/translate.py
chriswmackey: Added it as an option.

honeybee_radiance/cli/translate.py
chriswmackey: I think the strength of this method is that it gives you the radiance string in stdout so that the end user can decide what to do with it. You're right that it's weird to have radinace strings in a JSON, though, I'll just join the material sting and the geometry string into one to keep things consistent.

honeybee_radiance/cli/translate.py
chriswmackey: So are you saying that we should change all of the '--log-file' options across all of the CLI commands to be called `--output-file` instead?  If so, we can open an issue for it and I'll make the change when I can.

mostaphaRoudsari: Yes. and since this is a breaking change I think we should do it sooner than later.

mostaphaRoudsari: https://github.com/ladybug-tools/honeybee-radiance/issues/121

chriswmackey: Understood. I will open similar issues on the honeybee-energy and dragonfly repos.

Issue #118, opened 2020-05-19
Add CLI commands to get objects from the standards lib


chriswmackey:
These commands would parallel those over on honeybee-energy:
https://github.com/ladybug-tools/honeybee-energy/blob/master/honeybee_energy/cli/lib.py

And I know that @mingbopeng will need these commands in order to ass radiance properties to the Rhino plugin.

Issue #89, opened 2020-05-04
Add the ability to load non-abridged ModifierSets to from the standards library


chriswmackey:
We did something similar for ConstructionSets over in honeybee-energy and it has been generally useful to enable users to extend their libraries. We should do it here too.

PR #126, opened 2020-05-20
feat(cli): Add commands to retrieve modifiers from the lib

chriswmackey:
This should give @mingbopeng and Konrad what they need to assign default modifiers to the model.

Resolves https://github.com/ladybug-tools/honeybee-radiance/issues/89

Resolves https://github.com/ladybug-tools/honeybee-radiance/issues/118

### DIFF
--- a/honeybee_radiance/cli/lib.py
+++ b/honeybee_radiance/cli/lib.py
@@ -26,7 +26,7 @@ def lib():
 @lib.command('modifiers')
 @click.option('--output-file', help='Optional file to output the JSON string of '
               'the object. By default, it will be printed out to stdout',
-              type=click.File('w'), default='-')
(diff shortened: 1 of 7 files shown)


honeybee_radiance/cli/lib.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+@click.group(help='Commands for retrieving objects from the standards library.')
+def lib():
+    pass
+
+@lib.command('modifiers')
+@click.option('--output-file', help='Optional file to output the JSON string of '
+              'the object. By default, it will be printed out to stdout',
+              type=click.File('w'), default='-')
```
mostaphaRoudsari, line 29: For all the `options` I would set show_default to True so it shows up in help.

PR #181, opened 2020-07-04
build(docker): set non-root user and reduce image size

AntoineDao:
These changes to the Dockerfile result in the following outcomes:

* The default user on the container is not a root user
* Use `ubuntu:eon` (~79MB) as a base instead of `python:3.7` (~920MB) to drastically reduce image size (resulting image is ~280MB)
* Change final `WORKDIR` from `/opt/run-folder/` to `/home/ladybugbot/run` to match `honeybee-energy` image. This means we will have to also change the workdir on the operator once we change the docker image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,40 @@
-FROM python:3.7
+FROM ubuntu:eoan
 
-WORKDIR /tmp/
+MAINTAINER Ladybug Tools info@ladybug.tools
+
+# Install core software deps
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y \
+    curl \
+    ca-certificates \
+    python3.7 \
+    python3-pip \
+    unzip \
+    git \
+    && rm -rf /var/lib/apt/lists/*
+
+# Create non-root user
+RUN adduser ladybugbot --uid 1000 --disabled-password --gecos ""
+USER ladybugbot
+WORKDIR /home/ladybugbot
+
+# Install radiance
+ENV RAYPATH=/home/ladybugbot/lib
+ENV PATH="/home/ladybugbot/bin:${PATH}"
 RUN curl -L https://ladybug-tools-releases.nyc3.digitaloceanspaces.com/Radiance_5.5.2_Linux.zip --output radiance.zip \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,17 +1,40 @@
...
-ENV RAYPATH=/usr/local/lib
+# Create non-root user
+RUN adduser ladybugbot --uid 1000 --disabled-password --gecos ""
+USER ladybugbot
+WORKDIR /home/ladybugbot
 
-WORKDIR /usr/app/
+# Install radiance
+ENV RAYPATH=/home/ladybugbot/lib
+ENV PATH="/home/ladybugbot/bin:${PATH}"
```
mostaphaRoudsari, line 23: Is this change because of potential permission issues? I liked the older version which is the common way that Radiance users are used to but I understand if we have to do this differently because of the permissions.

Dockerfile
```diff
@@ -1,17 +1,40 @@
-FROM python:3.7
+FROM ubuntu:eoan
```
mostaphaRoudsari, line 1: Did you test running a Radiance command? I wonder if the Radiance build will work fine even if it has been built against a different version of ubuntu.

Dockerfile
AntoineDao: The `RAYPATH` var is just set so that radiance knows where to find the lib folder. It does differ from the "out of the box" install style but means that there is less permissions manipulation (running `chmod` commands) with the `/usr/local` folders.

Line 23 however is used to add `/home/ladybugbot/bin` to the known executables path so that if you type `> rtrace` the shell knows to look in the `/home/ladybugbot/bin` folder for it. Installing user scoped binaries in the `/home/<user-name>/bin` is an accepted practice for linux environments.

Dockerfile
AntoineDao: `rtrace` worked fine for me. I don't foresee any issues between ubuntu versions. Had it been compiled for a completely different arch like windows or macos I would be more concerned.

In principle it should work and my small tests indicate it does. We can carry out more tests by running workflows to determine whether the image base is suitable.

PR #186, opened 2020-07-05
Add octree command to CLI

mostaphaRoudsari:
This PR adds ocree command to CLI that works with honeybee Radiance folder.

This is the first CLI command that needs Radiance to execute successfully. To address that I added Radiance installation to Travis CI. It might take a couple of iterations to get the CI work correctly.


mostaphaRoudsari:
@AntoineDao, if can have a look at this whenever you get a chance it will be really helpful. It can't find Radiance commands, in this case oconv. Either copying the files in failing or the PATH is not set correctly. I tried changing both but still no luck! 😐

chriswmackey:
FYI @mostaphaRoudsari ,
This is the default location where honeybee is expecting the radiance installation to be on Linux:
https://github.com/ladybug-tools/honeybee-radiance/blob/master/honeybee_radiance/config.py#L214-L216
It looks like you might have set up the travis machine to install it there but I figured I'd let you know just in case.

AntoineDao:
image

mostaphaRoudsari:
Perfect! Thanks. I will send a PR for the updated recipes shortly.

PR #220, opened 2020-08-13
feat(dictutil): Add module to re-serialize any object from dict

### LINKED ISSUE #172: File names in radiance folder
Hi @chriswmackey, the folder looks really nice. Am I correct that all the opaque geometry and all the aperture will always be written to a single file? In that case are you fine if we do not include the model name in the name and keep it as aperture.rad and envelope or opaque.rad?

### LINKED ISSUE #192: Specify a type for files input to CLI commands
It looks like the best practice would be to follow something similar to what @AntoineDao implemented here:
https://github.com/ladybug-tools/honeybee-energy/pull/273
This will give us some extra checks if the user inputs something incorrect.

chriswmackey:
This commit adds a dictutil module to the package, with a dict_to_object method that can re-serialize any dictionary produced from the library back into a honeybee-radiance Python object.  We have implemented similar methods in honeybee-core and honeybee-energy and they have been very useful for keep in the JSON re-serialization code clean on the plugin side of things.

Given that the putil module already handles a lot of these re-serialization issues across honeybee-radiance, I didn't need to add much code in order to enable a package-wide re-serialization function.  I only needed to check the case of a few key top-level assets (ModifierSet, SensorGrid, View) and add a module to re-serialize all of the lightsource objects.

Once we merge this in, I will add the radiance re-serialization functionality into the Grasshopper plugin.

@mostaphaRoudsari , If you can take a quick look over the changes, it's always helpful. But there's nothing here that is outside the norm of what's been implemented on the other packages.

I'm also taking the opportunity of this PR to close a few small issues that were open:

Resolves #172
Resolves #192

### DIFF
--- a/honeybee_radiance/dictutil.py
+++ b/honeybee_radiance/dictutil.py
@@ -1,0 +1,50 @@
+# coding=utf-8
+"""Utilities to convertint any dictionary to Python objects.
+
+Note that importing this module will import almost all modules within the
+library in order to be able to re-serialize almost any dictionary produced
+from the library.
+"""
+from honeybee_radiance.putil import dict_to_primitive
+from honeybee_radiance.modifierset import ModifierSet
+from honeybee_radiance.sensorgrid import SensorGrid
+from honeybee_radiance.view import View
+from honeybee_radiance.lightsource.dictutil import dict_to_light_source, \
+    LIGHT_SOURCE_TYPES
+
+
+def dict_to_object(honeybee_radiance_dict, raise_exception=True):
+    """Re-serialize a dictionary of almost any object within honeybee_radiance.
+
+    This includes any Modifier, ModifierSet, LightSource, SensorGrid or View object.
+
+    Args:
+        honeybee_radiance_dict: A dictionary of any Honeybee radiance object. Note
+            that this should be a non-abridged dictionary to be valid.
+        raise_exception: Boolean to note whether an excpetion should be raised
+            if the object is not identified as a part of honeybee_radiance.
+            Default: True.
+
+    Returns:
+        A Python object derived from the input honeybee_radiance_dict.
+    """
+    try:  # get the type key from the dictionary
+        obj_type = honeybee_radiance_dict['type']
+    except KeyError:
+        raise ValueError('Honeybee_radiance dictionary lacks required "type" key.')
+
+    if obj_type == 'ModifierSet':
+        return ModifierSet.from_dict(honeybee_radiance_dict)
+    elif obj_type == 'SensorGrid':
+        return SensorGrid.from_dict(honeybee_radiance_dict)
+    elif obj_type == 'View':
+        return View.from_dict(honeybee_radiance_dict)
+    elif obj_type in LIGHT_SOURCE_TYPES:
+        return dict_to_light_source(honeybee_radiance_dict)
+    else:
+        try:
+            return dict_to_primitive(honeybee_radiance_dict)
+        except ValueError:
+            if raise_exception:
+                raise ValueError(
+                    '{} is not a recognized honeybee radiance object'.format(obj_type))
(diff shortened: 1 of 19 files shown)


honeybee_radiance/lightsource/dictutil.py
```diff
@@ -0,0 +1,56 @@
...
+from honeybee_radiance.lightsource.sky.climatebased import ClimateBased
+from honeybee_radiance.lightsource.sky.hemisphere import Hemisphere
+from honeybee_radiance.lightsource.sky.skydome import Skydome
+from honeybee_radiance.lightsource.sky.skymatrix import SkyMatrix
+from honeybee_radiance.lightsource.sky.sunmatrix import SunMatrix
+
+
+LIGHT_SOURCE_TYPES = \
+    ('Sunpath', 'Ground', 'CertainIrradiance', 'CIE', 'ClimateBased',
+     'Hemisphere', 'SkyDome', 'SkyMatrix', 'SunMatrix')
```
mostaphaRoudsari, line 16: What if we change this to a dictionary and use a quick look up instead of going through all the if / else to find the type. This should also keep the performance the same even if we add so many new types:

```python
LIGHT_SOURCE_TYPES = \
    {'Sunpath': Sunpath, 'Ground': Ground, 'CertainIrradiance': CertainIrradiance, 'CIE': CIE, 'ClimateBased': ClimateBased,
     'Hemisphere': Hemisphere, 'SkyDome': SkyDome, 'SkyMatrix': SkyMatrix, 'SunMatrix': SunMatrix}
```

Then for the function itself the code can be like this:

```python

def dict_to_light_source(light_source_dict, raise_exception=True):
    try:  # get the type key from the dictionary
        light_type = light_source_dict['type']
    except KeyError:
        raise ValueError('Load dictionary lacks required "type" key.')

    try:
        # I think this should work otherwise you can use getattr
        return LIGHT_SOURCE_TYPES[light_ype].from_dict(light_source_dict)
    except KeyError:
        raise ValueError(
            '{} is not a recognized radiance Light Source type'.format(light_type))
```

honeybee_radiance/dictutil.py
```diff
@@ -0,0 +1,50 @@
...
+
+    Returns:
+        A Python object derived from the input honeybee_radiance_dict.
+    """
+    try:  # get the type key from the dictionary
+        obj_type = honeybee_radiance_dict['type']
+    except KeyError:
+        raise ValueError('Honeybee_radiance dictionary lacks required "type" key.')
+
+    if obj_type == 'ModifierSet':
```
mostaphaRoudsari, line 36: I think this method can also be simplified similar to the comment that I left for lightsources. At least for the first three objects.

honeybee_radiance/lightsource/dictutil.py
chriswmackey: I would love to do that except that the name of the class you made is called `Skydome` while the type key is called `SkyDome`.  Would you want me to make the change to `SkyDome` as part of this PR? Then I can implement what you have here.

chriswmackey: Actually, I guess I can still get the dictionary to work that way since the key can be different from the value. Do you want me to make that change anyway so that we don't make mistakes like the one in your sample in the future?

mostaphaRoudsari: Yes. I just copy/pasted the keys. You can use the correct class names. Didn't really test the code.

honeybee_radiance/dictutil.py
chriswmackey: The strategy above definitely .makes sense for the lightsource dictutil module but I'm a little less inclined to do this here since it will most likely harm the performance given that it's much more likely that someone is re-serializing a primitive than  any of these 3 object. So the `except` case isn't really exceptional.  If we add many more classes for this module to handle, then I'm ok with changing it. But 3 doesn't seem like enough.

honeybee_radiance/lightsource/dictutil.py
chriswmackey: Done!

Issue #230, opened 2020-08-17
Add a property to SensorGrids class to optionally store mesh information


chriswmackey:
I have a feeling that this will help a lot with producing visualizations of grid-based results.

Issue #229, opened 2020-08-17
Add grids and views property to ModelRadianceProperties


chriswmackey:
This will make it easier to coordinate the grids and views with Models.

We should also ensure that the grids + views get moved, rotated, scaled whenever the Model is transformed or the units system is changed.

chriswmackey:
And we should ensure that the model.to.rad_folder method writes out the grids and views.

Issue #100, opened 2020-05-12
Add move, rotate, scale, and reflect commands to SensorGrid and View


chriswmackey:
This will be needed in order to coordinate these objects with transformed Models.

PR #231, opened 2020-08-18
feat(model): Add sensor_grid and views under ModelRadianceProperties

chriswmackey:
This commit adds several improvements around the SensorGrid and View classes:

* Include SensorGrids and Views as properties under ModelRadianceProperties
* Write SensorGrid and View files as part of the Model.to.rad_folder method
* Add ability to move, rotate, scale reflect sensor grids + views
* Ensure that the grids + views get moved, rotated, scaled whenever the Model is transformed or the units system is changed
* Added a property to the SensorGrid class to optionally store the mesh information for better visualization

Resolves https://github.com/ladybug-tools/honeybee-radiance/issues/230
Resolves https://github.com/ladybug-tools/honeybee-radiance/issues/229
Resolves https://github.com/ladybug-tools/honeybee-radiance/issues/100

### DIFF
--- a/honeybee_radiance/properties/model.py
+++ b/honeybee_radiance/properties/model.py
@@ -1,5 +1,7 @@
 # coding=utf-8
 """Model Radiance Properties."""
+from ..sensorgrid import SensorGrid
+from ..view import View
 from ..dynamic.group import DynamicShadeGroup, DynamicSubFaceGroup
 from ..modifierset import ModifierSet
 from ..mutil import dict_to_modifier  # imports all modifiers classes
@@ -23,6 +25,8 @@ class ModelRadianceProperties(object):
 
     Properties:
         * host
+        * sensor_grids
+        * views
         * modifiers
         * blk_modifiers
         * room_modifiers
@@ -34,16 +38,58 @@ class ModelRadianceProperties(object):
         * dynamic_subface_groups
         * shade_group_identifiers
         * subface_group_identifiers
+        * has_sensor_grids
+        * has_views
     """
 
-    def __init__(self, host):
+    def __init__(self, host, sensor_grids=None, views=None):
         """Initialize Model radiance properties."""
         self._host = host
+        self.sensor_grids = sensor_grids
+        self.views = views
 
     @property
     def host(self):
         """Get the Model object hosting these properties."""
         return self._host
+
+    @property
+    def sensor_grids(self):
+        """Get or set an array of SensorGrids that are associated with the model."""
+        return tuple(self._sensor_grids)
+
+    @sensor_grids.setter
+    def sensor_grids(self, value):
+        if value:
+            try:
+                self._sensor_grids = list(value)
+                for obj in self._sensor_grids:
+                    assert isinstance(obj, SensorGrid), 'Expected SensorGrid for Model' \
+                        ' sensor_grids. Got {}.'.format(type(value))
+            except (ValueError, TypeError):
+                raise TypeError(
+                    'Model sensor_grids must be an array. Got {}.'.format(type(value)))
+        else:
+            self._sensor_grids = []
+
+    @property
+    def views(self):
+        """Get or set an array of Views that are associated with the model."""
+        return tuple(self._views)
+
+    @views.setter
+    def views(self, value):
+        if value:
+            try:
+                self._views = list(value)
+                for obj in self._views:
+                    assert isinstance(obj, View), 'Expected View for Model' \
+                        ' views. Got {}.'.format(type(value))
+            except (ValueError, TypeError):
+                raise TypeError(
+                    'Model views must be an array. Got {}.'.format(type(value)))
+        else:
+            self._views = []
 
     @property
     def modifiers(self):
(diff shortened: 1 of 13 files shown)


honeybee_radiance/properties/model.py
```diff
@@ -34,17 +38,59 @@ class ModelRadianceProperties(object):
...
         return self._host
 
+    @property
+    def sensor_grids(self):
+        """Get or set an array of SensorGrids that are associated with the model."""
+        return tuple(self._sensor_grids)
+
+    @sensor_grids.setter
+    def sensor_grids(self, value):
+        if value is not None:
```
mostaphaRoudsari, line 63: We can probably change this to `if value:` to catch empty lists or tuples.

honeybee_radiance/properties/model.py
```diff
@@ -310,6 +467,72 @@ def check_duplicate_modifier_set_identifiers(self, raise_exception=True):
             return False
         return True
 
+    def check_duplicate_sensor_grid_identifiers(self, raise_exception=True):
```
mostaphaRoudsari, line 470: I had to resolve a similar issue before and this was the best option that I found online.

```py
dup = [t for t, c in collections.Counter(grid.identifier for grid in self.sensor_grids).items() if c > 1]
```

honeybee_radiance/sensor.py
```diff
@@ -92,9 +156,6 @@ def __eq__(self, value):
     def __ne__(self, value):
         return not self.__eq__(value)
 
-    def ToString(self):
```
mostaphaRoudsari, line 95: Why are we deleting `ToString` here?

honeybee_radiance/sensor.py
```diff
@@ -49,14 +51,76 @@ def from_raw_values(cls, x=0, y=0, z=0, dx=0, dy=0, dz=1):
...
+        self._dir = typing.tuple_with_length(value) if value is not None else (0, 0, 1)
+
+    def move(self, moving_vec):
+        """Move this sensor along a vector.
+
+        Args:
+            moving_vec: A ladybug_geometry Vector3D with the direction and distance
+                to move the sensor.
+        """
+        self.pos = tuple(pv.Point3D(*self.pos).move(moving_vec))
```
mostaphaRoudsari, line 77: What if we use `Point3D` and `Vector3D` instead of tuple now that you need it for all the transforms?

We can add a classmethod for creating sensors from tuples or list.

honeybee_radiance/sensorgrid.py
```diff
@@ -31,23 +32,20 @@ class SensorGrid(object):
...
-        self._sensors = tuple(sensors)
-        for sen in self._sensors:
-            if not isinstance(sen, Sensor):
-                raise ValueError(
-                    'Sensors in SensorGrid must be form type Sensor not %s' % type(sen)
-                )
+        self.sensors = sensors
         self._room_identifier = None
         self._light_path = None
+        self._mesh = None
```
mostaphaRoudsari, line 48: I feel we might want to add a `classmethod` to create a sensor grid from an input mesh. That will come handy in several cases.

honeybee_radiance/view.py
```diff
@@ -738,31 +738,90 @@ def to_file(self, folder, file_name=None, mkdir=False):
...
+        view_position = pv.Point3D(*self.position)
+        self.position = view_position.scale(factor, origin)
+        self.direction = pv.Vector3D(*self.direction) * factor
+        self.up_vector = pv.Vector3D(*self.up_vector) * factor
+
+    def _apply_plane_properties(self, plane, view_direction, view_up_vector):
+        """Re-set the position, direction and up_vector from a Plane.
+
+        This method also ensures that the magnitude of the vectors is unchanged
+        (since all Plane objects will have unitized vectors).
```
mostaphaRoudsari, line 820: Hmmm! I have to check but I think in Radiance the vector length doesn't make a difference and we can keep it always unitized.

honeybee_radiance/writer.py
```diff
@@ -322,6 +323,21 @@ def model_to_rad_folder(model, folder=None, folder_type=2, config_file=None,
         new_bsdf_path = os.path.join(bsdf_folder, bsdf_name)
         shutil.copy(bdf_mod.bsdf_file, new_bsdf_path)
 
+    # write the assigned sensor grids and views into the correct folder
+    grid_dir = model_folder.grid_folder(full=True)
+    for grid in model.properties.radiance.sensor_grids:
+        write_to_file_by_name(
+            grid_dir, '{}.pts'.format(grid.identifier), grid.to_radiance())
```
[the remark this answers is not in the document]
chriswmackey: I didn't realize the sensor grid has a `to_file` method.  I'll use that.

honeybee_radiance/properties/model.py
chriswmackey: I'm not totally convinced it's the most readable or beautiful but it's the shortest.  If we decide that we like it, there are some ~30 methods that should be changed across honeybee-core, dragonfly-core and the extensions.

honeybee_radiance/sensor.py
chriswmackey: This was a redefinition of `ToString`. You'll see that it already exists on the class if you scroll up.

honeybee_radiance/sensor.py
chriswmackey: Because you built in all of these radiance serialization capabilities for the class that rely on everything being a tuple.  If you want to re-do all of these radiance serialization functions to use the ladybug_geometry objects instead of tuples, maybe then it will make sense.  But Radiance serialization is a more common use case than transforms so, at the moment, it wasn't worth all of this effort.

honeybee_radiance/sensorgrid.py
chriswmackey: Good point. Will do.

honeybee_radiance/view.py
chriswmackey: If it didn't make a difference, then I imagine that you would not have the following docstring in the view class:
```
direction: Set the view direction (-vd) vector to (x, y, z). The
    length of this vector indicates the focal distance as needed by
    the pixel depth of field (-pd) in rpict. Default: (0, 0, 1)
```

honeybee_radiance/sensor.py
chriswmackey: Not to mention the fact that so many of the tests would need to be changed.  It just seems like a task where, if I were to do it, there's a very high probability of breaking things.

honeybee_radiance/view.py
mostaphaRoudsari: That's from Radiance's documentation and you are right.

honeybee_radiance/sensor.py
mostaphaRoudsari: OK. then let's leave it as is.

honeybee_radiance/properties/model.py
mostaphaRoudsari: In that case, we should probably make one method in honeybee-core and have every other library to use it. It is not critical of course as it will work as is.

honeybee_radiance/sensorgrid.py
chriswmackey: Done

chriswmackey:
@mostaphaRoudsari ,
Some of the things you brought up here seem like we should probably open up separate issues about them. I re-requested your review to see if you agree.

honeybee_radiance/sensor.py
chriswmackey: On second though, maybe I could do it easily for the SensorGrid if you can take care of it for the View class.  I just don't know how we should handle all of these properties that are `TupleOption`.

honeybee_radiance/properties/model.py
chriswmackey: Ok. I opened an issue for it https://github.com/ladybug-tools/honeybee-core/issues/198

PR #411, opened 2021-01-26
add support for sensor grid groups

mostaphaRoudsari:
This commit adds support for grouping sensor groups together.

### DIFF
--- a/honeybee_radiance/sensorgrid.py
+++ b/honeybee_radiance/sensorgrid.py
@@ -23,7 +23,9 @@ class SensorGrid(object):
     Args:
         identifier: Text string for a unique SensorGrid ID. Must not contain spaces
             or special characters. This will be used to identify the object in the
-            exported Radiance files.
+            exported Radiance files. You may use / in name to identify grid groups.
+            For example floor_1/living_room create a sensor grid with living_room
+            identifier and floor_1 group identifier.
         sensors: A collection of Sensors.
 
     Properties:
@@ -36,14 +38,17 @@ class SensorGrid(object):
         * light_path
         * mesh
         * base_geometry
+        * group_identifier
+
     """
 
     __slots__ = ('_identifier', '_display_name', '_sensors', '_room_identifier',
-                 '_light_path', '_mesh', '_base_geometry')
+                 '_light_path', '_mesh', '_base_geometry', '_group_identifier')
 
     def __init__(self, identifier, sensors):
         """Initialize a SensorGrid."""
-        self.identifier = typing.valid_rad_string(identifier)
+        self._group_identifier = None
+        self.identifier = identifier
         self._display_name = None
         self.sensors = sensors
         self._room_identifier = None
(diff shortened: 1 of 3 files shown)


honeybee_radiance/sensorgrid.py
```diff
@@ -23,7 +23,9 @@ class SensorGrid(object):
     Args:
         identifier: Text string for a unique SensorGrid ID. Must not contain spaces
             or special characters. This will be used to identify the object in the
-            exported Radiance files.
+            exported Radiance files. You may use / in name to identify grid groups.
+            For example floor_1/living_room create a sensor grid with living_room
+            identifier and floor_1 group identifier.
```
chriswmackey, line 28: Something this implicit is definitely ok for the Grasshopper interface but I was imagining that the core libraries would be more explicit about making you set the `group_identifier` if it's something that you want. For example, I'm not sure that I'm a fan of the idiosyncrasy that it creates in the dict schema because you can technically set the `group_identifier` two ways from a dictionary (one that explicitly uses a key and another that just includes a `/` in the identifier).

I can also see a case at some point where someone includes a `/` in their input for the sensorgrid identifier, they don't get an error as they usually would would for illegal radiance characters in IDs and so they think `/` is a legal radiance character (like it is in EnergyPlus), Then, they go to select sensorgrids from a model using the identifer that they originally input to the sensor grid here and they get nothing back because the actual grid identifier doesn't have a `/` in it.

I guess this just has a code smell to it.

Maybe it also helps to know that this implicit feature is not something that I am going to use on the Grasshopper interface because the generic methods that process the user's `_name_` input will clean out the `/` before assigning it as an identifier. So I'm going to have to write a small custom method for this case on the interface end anyway and I guess the only use case this serves is to save Python scripters a single line of code for having to explicitly set the group identifier. So I'm not really sure it's worth it.

honeybee_radiance/sensorgrid.py
```diff
@@ -283,6 +296,22 @@ def room_identifier(self):
...
+        This will be used in the write to radiance folder method to write all the grids
+        with the same group identifier under the same subfolder.
+
+        If None, the grid will be written to the root of grids folder.
+        """
+        return self._group_identifier
+
+    @group_identifier.setter
+    def group_identifier(self, n):
+        identifier_key = '/'.join(typing.valid_string(key) for key in n.split('/'))
```
chriswmackey, line 312: You should use `valid_rad_string` here instead of `valid_string`.  The former will only check for the illegal radiance characters and not the eillegal characters for both E+ and Radiance.

I think the ability to create sub-sub-folders by using multiple `/` is a good one but I think you should make a note of this capability in the `group_identifier` docstring.

chriswmackey (changes requested):
This is generally ok, @mostaphaRoudsari , but the implicitness of the `identifier.setter` behavior has a code smell to me.  Would you mind changing it before we merge?

I promise that I will make the `/` behavior on the Grasshopper interface will work as you intend it.

mostaphaRoudsari:
@chriswmackey, I removed the option to pass it as part of the name.

chriswmackey (approved):
LGTM!

PR #414, opened 2021-01-29
fix(sensorgrid): Remove argument that flips the grids by default

devngc:


devngc:
@chriswmackey, I believe this is the reason the grids that I am creating from Face3D are being flipped in honeybee-3dm. 
resolves https://github.com/ladybug-tools/honeybee-3dm/issues/116

chriswmackey (changes requested):
Thanks, @devngc .

I agree that a change is in order here but, rather than deleting this argument, I think it would be best to expose the `flip` argument that's going to the ladybug_geometry `Face3D.mesh_grid` method on this `SensorGrid.from_face3d` classmethod here. I'm sure there are cases where we would want to generate a flipped grid, specifically when generating grids from floor faces of Rooms (which always point downwards).

I verified that this method isn't being used anywhere in the Grasshopper plugin right now so you should feel free to make the default to not flip the mesh.

chriswmackey (approved):
LGTM!  I'll merge this now so that it makes it into the update of the honeybee-radiance docker image on Pollination that @mostaphaRoudsari 's doing now.

PR #450, opened 2021-02-24
fix(docker): reduce image size

tymokvo:
# What
- Reduce image size

# Why
- To enable faster pulls

# Notes
@mostaphaRoudsari I see that this is built in `deploy.sh` but I'm not sure where in the CI process this gets called so I'm not positive that `curl` is available. As long as it is, it should be fine.

The changes here reduce uncompressed image size `1.02GB` -> `0.33GB`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,9 @@ LABEL maintainer="Ladybug Tools" email="info@ladybug.tools"
 ARG radiance_version
 
 ENV WORKDIR='/home/ladybugbot'
-ENV RAYPATH='${WORKDIR}/lib'
+ENV RUNDIR="${WORKDIR}/run"
+ENV LIBRARYDIR="${WORKDIR}/honeybee-radiance"
+ENV RAYPATH="${WORKDIR}/lib"
 ENV PATH="${WORKDIR}/.local/bin:${RAYPATH}:${PATH}"
 
 RUN apt-get update \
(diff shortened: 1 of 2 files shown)


mostaphaRoudsari (approved):
Hi @tymokvo, looks good and it should work.

> I'm not sure where in the CI process this gets called so I'm not positive that curl is available. As long as it is, it should be fine.

It's called as part of the semantic release (https://github.com/ladybug-tools/honeybee-radiance/blob/d3a69de4325ce3ce9e823efba083bc36dd69a3bd/.github/workflows/ci.yaml#L70). semantic release uses the `.releaserc.json` which is where the plugins are set. The last one `@semantic-release/exec` executes the commands when there is a new release which is where deploy.sh is called: (https://github.com/ladybug-tools/honeybee-radiance/blob/d3a69de4325ce3ce9e823efba083bc36dd69a3bd/.releaserc.json#L9). [Here](https://github.com/semantic-release/exec#semantic-releaseexec) is a documentation for `exec` plugin and [here is a list of other plugins](https://semantic-release.gitbook.io/semantic-release/extending/plugins-list) just in case.

mostaphaRoudsari:
Hi @tymokvo, I'm going to merge this one on your behalf to both see if it works as expected and to get you an updated Pollination recipe to test by the time that you start your day. 😎

mostaphaRoudsari:
Ah! I didn't notice that the commit messages are not formatted and you meant to squash and merge these. I send another PR with post-processing functionalities that should trigger a new release.

I suggest to use [commitizen](https://github.com/commitizen/cz-cli) for your commits so you won't forget about formatting them. We should probably also enforce it by adding the configs to the repositories itself.

tymokvo:
Yep! Sorry, should have said something. I have commitizen configured but I find it really slows me down so I usually just squash the feature branch commits into one formatted one for the PR.

PR #458, opened 2021-02-26
fix(docker): add radiance binaries to PATH

tymokvo:
Add the ol' binaries to the ol' path.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,9 +5,9 @@ LABEL maintainer="Ladybug Tools" email="info@ladybug.tools"
 ENV WORKDIR='/home/ladybugbot'
 ENV RUNDIR="${WORKDIR}/run"
 ENV LIBRARYDIR="${WORKDIR}/honeybee-radiance"
-ENV RAYPATH="${WORKDIR}/lib"
+ENV RAYPATH=".:${WORKDIR}/lib"
 ENV BINPATH="${WORKDIR}/bin"
-ENV PATH="${WORKDIR}/.local/bin:${BINPATH}:${RAYPATH}:${PATH}"
+ENV PATH="${WORKDIR}/.local/bin:${BINPATH}:${PATH}"
 
 RUN apt-get update \
     && apt-get -y install --no-install-recommends git \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -6,7 +6,8 @@ ENV WORKDIR='/home/ladybugbot'
 ENV RUNDIR="${WORKDIR}/run"
 ENV LIBRARYDIR="${WORKDIR}/honeybee-radiance"
 ENV RAYPATH="${WORKDIR}/lib"
```
mostaphaRoudsari, line 8:
```suggestion
ENV RAYPATH=".:${WORKDIR}/lib"
```

Dockerfile
```diff
@@ -19,8 +20,8 @@ USER ladybugbot
 WORKDIR ${WORKDIR}
 
 # Expects a decompressed radiance folder in the build context
-COPY radiance/usr/local/radiance/bin ${WORKDIR}/bin
-COPY radiance/usr/local/radiance/lib ${WORKDIR}/lib
+COPY radiance/usr/local/radiance/bin ${BINPATH}
+COPY radiance/usr/local/radiance/lib ${RAYPATH}
```
mostaphaRoudsari, line 24: Sorry for making it not as good looking by adding the current directory to RAYPATH.

```suggestion
COPY radiance/usr/local/radiance/lib ${WORKDIR}/lib
```

tymokvo:
:arrow_up: These make much more sense together. Thank you @mostaphaRoudsari !

Dockerfile
```diff
@@ -19,8 +20,8 @@ USER ladybugbot
...
 
 # Install honeybee-radiance
 COPY honeybee_radiance ${LIBRARYDIR}/honeybee_radiance
```
mostaphaRoudsari, line 27: @tymokvo, one more comment here! Does this compile the entrypoints for running honeybee-radiance commands from the command line? Don't we need to use pip install to create the executable files under Scripts folder?

I'm lazy to build the docker and test it myself but what happens when you run `honeybee-radiance --help` from the command line? If it gives you the right output then ignore my comment here.

mostaphaRoudsari (commented):
Sorry! Added one more comment.

Dockerfile
tymokvo: Yep, the cli help prints out. I believe `pip` handles that installation on line 38.

Dockerfile
tymokvo: It's ok :crying_cat_face:

tymokvo:
I thought from your description that `RAYPATH` needed to be removed from `PATH` so I added a new commit instead of accepting the suggestions. Even though they seem perfectly good to me!

Dockerfile
mostaphaRoudsari: You are right. Didn't see the existing lines on my phone. :|

mostaphaRoudsari (approved):
Looks great!

PR #463, opened 2021-03-01
docs(cli): Improve CLI documentation structure

santiagogaray:
Hi @chriswmackey,

Here is my first pass to this library. This one has a a few unique things and thought would be a good one to start with before we move on to the rest of the libraries. 

1.  **CLI group names:**
The main issue here is that there are two CLI groups (mtx and postprocess) where the filenames differ from the group names. To address this I created a hash table where each library can include `rst_filename: group_name` pairs for these special cases. This means that if future CLI groups fall in this special case they will need to be added to the hash table for the group documentation to be generated properly. If we don't want to rely on people adding them to the hash table manually we could possibly automate this by reading the file with the CLI code and extract the name of the group from there. May be overkill to do this for every group documented on every library, specially if this is only useful for a handful of times so clearly there is a trade off to consider with this approach. Hope this is clear and happy to show the code if you need more details.

2.  **Installation section:** 
I have included the Installation section inside the CLI Docs to be consistent with what we used in honeybee-energy. I just copied the information that was include in the main page. If there is any other information we want to include here, please let me know, happy to add it!

3. **Warnings during documentation build:**
A couple of warnings we can fix here or in a future PR:
- honeybee-radiance\docs\cli\mtx.rst. From what I gathered this seems to be triggered somehow inside the mtx.py file by the `/` and  `*` options within the `@click.options` decorator. Seems like these special characters are creating some conflicts. The rendered html is showing these two options in link format. Tried escaping them ('\\') but the forward slash is still giving me some trouble, also not sure this escaping affects the code or not.  Please let me know what do you think, I can look deeper into it if the result is not acceptable.

- honeybee-radiance lightsource\sky\CIE.py . These are some some lists formatting issues, can be easily corrected but not sure I should touch this file in this PR.
 
Sorry @chriswmackey for the long list in something that should be pretty straightforward, the good news is that the other two libraries I updated are issue-less! :)

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -1,6 +1,0 @@
-Honeybee Radiance Command Line Interface
-========================================
-
-.. click:: honeybee_radiance.cli:radiance
-   :prog: honeybee_radiance
-   :show-nested:
--- a/docs/cli/index.rst
+++ b/docs/cli/index.rst
@@ -1,0 +1,7 @@
+CLI Docs
+========
+
+Installation
+------------
+
+To check if the command line interface is installed correctly use `honeybee-radiance --help`.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -229,6 +229,250 @@ autodoc_default_options = {
 
 autodoc_member_order = 'groupwise'
 
+# -- CLI documentation  -------------------------------------------------------
+"""Improves the CLI documentation section.
+
+In order to have separate html pages for each module inside library\\cli
+additional documentation reST files(.rst) need to be generated in docs\\cli folder.
+
+Note:
+    This process assumes that each cli module represent a group of subcommands.
+    Use hash tables below to address special cases.
+"""
+# Activate this CLI documentation process
+custom_cli_docs = True
+
+# Repository/library hash table.
+# If respository not found in table, the module name will be extracted from the
+# modules.rst file in the project folder.
+# Format: {repository_name: library_name}
+ht_repo_lib = {}
+
+# Library/command hash table.
+# Created to address CLI tool names that differ from their library name beyond
+# the underscore(_) to dash(-) difference.
+# Format: {library_name: tool_name}
+ht_lib_tool = {}
+
+# Cli_file/cli_group hash table.
+# Created to address CLI file names that differ from the included CLI group name
+# Format: {
+#    library_name_1: {CLI_filename_1: group_name_1,
+#                     CLI_filename_2: group_name_2
+#                    },
+#    library_name_2: {CLI_filename_3: group_name_3,
+#                     CLI_filename_4: group_name_4
+#                    }
+# }
+ht_cli_file_group = {
+    'honeybee_radiance': {'mtx': 'mtxop',
+                          'postprocess': 'post_process'
+                          }
+}
+
+
+# Specify if existing reSt files found in docs//cli folder will be overwritten.
+cli_overwrite = False
+
+
+def create_cli_files():
+    """Generate additional files required by sphinx to build well structured
+     CLI documentation pages.
+
+    The process consists in generating reST files with sphinx-click directives
+    for each click group detected in the library's cli folder and updating
+    the index.rst file to include a list of the click groups found.
+    """
+
+    # Get CLI data from library
+    proj_folder = os.path.dirname(__file__)
+    result = get_cli_data(proj_folder)
+    if not result:
+        return
+
+    all_groups, lib_name, tool_name = result
+
+    # Prepare docs folder and reST files to create
+    doc_folder = os.path.join(proj_folder, 'cli')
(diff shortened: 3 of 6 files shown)


chriswmackey:
Thanks @santiagogaray ,

Your changes generally look good and the high-level summary of questions is very helpful. To go down the list:

1. I didn't realize that we had cases where the CLI group didn't match the module name and we should get @mostaphaRoudsari 's thoughts on why this is the case here. I'm pretty sure that we do not want to change the name of the CLI group since this would mess up a lot of things that are using the CLI but we may want to change the module names to match the CLI group. I don't know if there is any reason why `mtx` is the module name yet `mxtop` is the group name but I feel this is probably one that we should change unless @mostaphaRoudsari has some reason for it. I understand why the `post_process` module does not match and I know that Mostapha is more of a fan than I of putting `_` into module names than I am.  Now that I see a case like this, I would agree that the module should be changed to be called `post_process` instead of `postprocess`. Or, I guess I would have called the CLI group name `postprocess` to begin with since postprocess is a word in the dictionary but it seems too much of a hassle to go through that change now. In any case, I think it's important for the names to match and be consistent and we should change the module names.  It just makes the code less readable/understandable when they aren't matched. Do you agree @mostaphaRoudsari ?

2. Including the installation info sounds good. Admittedly, it's a little less relevant now that we got rid of the `[cli]` extra but I think it's still helpful.

3. I am not completely sure of the best solution for the `\` and `*` choices here. Possibly, if we put the prefix `r` in front of the choice (eg. `r'*'`) to make it a literal string, click might know how to deal with it correctly. If not, maybe we just leave this issue for later.

4. You should definitely feel free to edit the doctrsings for the cie.py module. I see that some of them currently don't display correctly and it would be nice to have them fixed. Any reformatting that you want to do of the docstrings is fair game.

mostaphaRoudsari:
Hi @chriswmackey and @santiagogaray,

For `post-process` I would have liked to call the module `post_process` but I wanted to keep the module names in honeybee consistent. I know @chriswmackey has strong opinion on not using `_` in module names so I kept it out.

For `mtx` I think `mtx` is the better name but radiance command is called `mtxop`. I followed that naming convention to keep it familiar for radiance users.


I'm fine renaming the modules to match the command group but I think we shouldn't. I would consider this a bug in the documentation plugin. If click allows for overwriting the name to provide a better interface then the documentation plugin should do the same. Did you try to reach out to the developers of click-cli documentation and get their input. It might be an easy fix on their side. Maybe we can even help with that and submit a PR to fix this.,

santiagogaray:
Hi @mostaphaRoudsari and  @chriswmackey,

I just want to clarify how we ended up with this "bug" in the cli documentation. This issue is on us and not on the click-cli documentation plug-in since at this point we are creating a workaround to get a better structured CLI doc (due to, yes, a big limitation in the plug-in).  
To briefly describe the issue, In order to create a better structured documentation I was left with a workaround adding a routine in conf.py that creates our own rst files for every group, using the CLI module filename in the directive options. In this directive you can place different module name and command group names, but I assumed I did not have to retrieve group names since they where the same. And that can be solved in a couple different ways as I explained above.
So in few words, the limitation in the pug-in is the lack of structuring options not the overriding of command group name. I think though our solution (with the group fix) should work just fine, regardless of our approach with the library naming.

mostaphaRoudsari:
Hi @santiagogaray, thank you for the clarification. In that case, I think this is something that we should be able to handle on our side. Since this is a feature offered by click I imagine we might be using it in other libraries too and we should be able to handle it IMO.

chriswmackey:
Thanks for your input, @mostaphaRoudsari , and for the explanation, @santiagogaray . I know that the workaround isn't really because of a click bug but it's just because we have a larger-than-average CLI and so we made the stylistic decision to break down the documentation with one web page per CLI group.

Now that I see Mostapha has reasons for making the module name different than the CLI group name, let's just keep your current fix for this, Santiago . I think we can merge whenever you are ready and I will leave the decision to you if you want to clean up the cie.py module or try using literal strings for the choice arguments before we merge.

santiagogaray:
Sounds good @chriswmackey. I did minor style fixes to the routine and left it with the hash table solution as you indicated. 

Regarding the mtx.py warnings, the r'*' did not fix it unfortunately, and no matter what I tried, could not make those warnings go away without triggering other warnings. Not understanding what's really going on behind the hood with sphinx, I decide to leave it alone, and have this minor issue in the output html.

I cleaned up the cie.py docstrings warnings (plus one minor docstring on _skybase.py).

Unless you guys have other comments, this PR should be good to merge.

chriswmackey:
Thanks, @santiagogaray , This all looks good to me and I will merge it now.

chriswmackey:
They look beautiful!
https://www.ladybug.tools/honeybee-radiance/docs/cli/index.html

PR #495, opened 2021-03-22
Always write grid folder with _info.json

### LINKED ISSUE #494: Always write `grid` folder with `_info.json` file
We should always write the `grid` folder and include the `_info.json` file even if there are no grid in the model. This makes the folder API a bit more consistent and avoids some bugs in the honeybee-radiance Pollination Plugin.

https://github.com/ladybug-tools/honeybee-radiance/blob/fae35ff1200f8a935e373d79ce38e9bdd4aa71da/honeybee_radiance/writer.py#L377-L397

AntoineDao:
fix #494

I also reproduced the `write empty folder is model does not contain property` behaviour for BSDF materials and dynamic properties. I am not sure how useful the empty BSDF folder is but having an empty `state.json` for dynamic properties makes sense as I assume the file is used in a similar manner to the `_info.json`

### DIFF
--- a/honeybee_radiance/writer.py
+++ b/honeybee_radiance/writer.py
@@ -294,8 +294,8 @@ def model_to_rad_folder(
     ext_dict = {}
     out_subfolder = model_folder.aperture_group_folder(full=True)
     dyn_subface = model.properties.radiance.dynamic_subface_groups
+    preparedir(out_subfolder)
     if len(dyn_subface) != 0:
-        preparedir(out_subfolder)
         for group in dyn_subface:
             if group.is_indoor:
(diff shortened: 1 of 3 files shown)


mostaphaRoudsari (approved):
We probably have to revisit this soon when we get to view-based recipes but until then this will provide what we need. Thanks!

I'll merge this in so we can push a new release for all the daylighting recipes.

chriswmackey:
Ah, I just thought of a much better way to handle this. We should just add options to the `model-to-rad` folder to `--check-grids` and `--check-views` and we can raise a clear exception errors when they are missing from the relevant recipes.  Can we just revert this and I'll send a PR with the new fix?

PR #499, opened 2021-03-25
fix(writer): Replace writing of empty folders with explicit exception

chriswmackey:
This commit adds options to perform checks for the presence of grids or views and so that an explicit exception can be raised for recipes requiring these inputs. This seems better than writing empty folders and failing silently.

It also seems more pythonic and elegant for two reasons:
- Explicit is better than implicit (it's much better to get an error about what's wrong than a silent failure)
- Beautiful is better than ugly (why should I get a bunch of empty folders when they are not needed).

### DIFF
--- a/honeybee_radiance/cli/translate.py
+++ b/honeybee_radiance/cli/translate.py
@@ -43,10 +43,19 @@ def translate():
 @click.option('--minimal/--maximal', help='Flag to note whether the radiance strings '
               'should be written in a minimal format (with spaces instead of line '
               'breaks).', default=False, show_default=True)
+@click.option('--no-grid-check/--grid-check', ' /-gc', help='Flag to note whether the '
+              'model should be checked for the presence of sensor grids. If the check '
+              'is set and the model has no grids, an explicit error will be raised.',
+              default=True, show_default=True)
+@click.option('--no-view-check/--view-check', ' /-vc', help='Flag to note whether the '
+              'model should be checked for the presence of views. If the check '
+              'is set and the model has no views, an explicit error will be raised.',
+              default=True, show_default=True)
 @click.option('--log-file', help='Optional log file to output the path of the radiance '
               'folder generated from the model. By default this will be printed '
               'to stdout', type=click.File('w'), default='-')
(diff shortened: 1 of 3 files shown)


mostaphaRoudsari (commented):
Hey @chriswmackey, I agree that this is a more elegant approach but if we want to make this change we have to update the function in honeybee-radiance and then update all the recipes. The current solution will work as is and we only need to update the dependency version.

I will not be able to take care of those changes tonight but if you can wait until later tomorrow or make the changes yourself then feel free to merge this in.

honeybee_radiance/cli/translate.py
```diff
@@ -58,8 +67,14 @@ def model_to_rad_folder(model_json, folder, view, grid, config_file, minimal, lo
         if folder is None:
             folder = os.path.dirname(os.path.abspath(model_json))
 
-        # re-serialize the Model and translate the model to a radiance folder
+        # re-serialize the Model and perform any checks
         model = Model.from_file(model_json)
+        if not no_grid_check and len(model.properties.radiance.sensor_grids) == 0:
```
mostaphaRoudsari, line 72: Does the model always have `properties.radiance` attribute? I imagine you can have cases where the model is generated without honeybee-radiance being installed and this attribute will be missing.

You may want to add a check for `hasattr(model.properties, 'radiance')` before checking the length of sensor grids. Same for the view part.

mostaphaRoudsari (changes requested):
See my comment about checking the attr.

honeybee_radiance/cli/translate.py
chriswmackey: As long as `honeybee-radiance` is installed, all Python model objects will have `properties.radiance`.  If there were no radiance properties in the Model JSON that was loaded, then all of the radiance properties are the default.

So there's no need for an extra `hasattr` check here.

chriswmackey:
Thanks for the quick review, @mostaphaRoudsari . I was planning to update the function and all of the relevant recipes tomorrow as part of the whole update of pollination-dsl. I would much rather "do it right" now rather than wait till later when there may be several more radiance recipes that need to be updated.

PR #511, opened 2021-04-01
feat(cli): add rpict

mostaphaRoudsari:
This commit adds rpict to honeybee-radiance. We need [this PR](https://github.com/ladybug-tools/honeybee-radiance-command/pull/143) in honeybee-radiance-command to be merged in first, update the dependencies and then this should work.

cc: @devngc

### DIFF
--- a/honeybee_radiance/cli/rpict.py
+++ b/honeybee_radiance/cli/rpict.py
@@ -1,0 +1,77 @@
+"""honeybee radiance rpict command."""
+import click
+import sys
+import logging
+import os
+
+from honeybee_radiance.config import folders
+from honeybee_radiance_command.rpict import Rpict, RpictOptions
+
+
+_logger = logging.getLogger(__name__)
+
+
+@click.group(help='Commands to run rpict in Radiance.')
+def rpict():
+    pass
+
+
+@rpict.command('rpict')
+@click.argument(
+    'octree', type=click.Path(exists=True, file_okay=True, resolve_path=True)
+)
+@click.argument(
+    'view', type=click.Path(exists=True, file_okay=True, resolve_path=True)
+)
+@click.option(
+    '--rad-params', help='Radiance parameters.'
+)
+@click.option(
+    '--rad-params-locked', help='Protected Radiance parameters. These values will '
+    'overwrite user input rad parameters.'
+)
+@click.option(
+    '--output', '-o', help='Path to output file. If a relative path is provided it '
+    'should be relative to project folder.'
+)
+@click.option(
+    '--dry-run', is_flag=True, default=False, show_default=True,
+    help='A flag to show the command without running it.'
+)
+def rpict_command(
+        octree, view, rad_params, rad_params_locked, output, dry_run):
+    """Run rpict command for an input octree and a view file.
+
+    \b
+    Args:
+        octree: Path to octree file.
+        view: Path to view file.
+
+    """
+    try:
+        options = RpictOptions()
+        # parse input radiance parameters
+        if rad_params:
+            options.update_from_string(rad_params.strip())
+        # overwrite input values with protected ones
+        if rad_params_locked:
+            options.update_from_string(rad_params_locked.strip())
+
+        # create command.
+        rpict = Rpict(
+            options=options, output=output, octree=octree, view=view
+        )
+
+        if dry_run:
+            click.echo(rpict)
+        else:
+            env = None
+            if folders.env != {}:
+                env = folders.env
+            env = dict(os.environ, **env) if env else None
+            rpict.run(env=env)
+    except Exception:
+        _logger.exception('Failed to run rpict command.')
+        sys.exit(1)
+    else:
+        sys.exit(0)


honeybee_radiance/cli/rpict.py
```diff
@@ -0,0 +1,74 @@
...
+
+from honeybee_radiance.config import folders
+from honeybee_radiance_command.rpict import Rpict, RpictOptions
+
+
+_logger = logging.getLogger(__name__)
+
+
+@click.group(help='Commands to run rpict in Radiance.')
+def raytrace():
```
devngc, line 15: def rpict():

honeybee_radiance/cli/rpict.py
```diff
@@ -0,0 +1,74 @@
...
+    pass
+
+
+@raytrace.command('rpict')
```
devngc, line 19: @rpict.command('rpict')

honeybee_radiance/cli/rpict.py
```diff
@@ -0,0 +1,74 @@
...
+    '--output', '-o', show_default=True, help='Path to output file. If a relative path'
+    ' is provided it should be relative to project folder.'
+)
+@click.option(
+    '--dry-run', is_flag=True, default=False, show_default=True,
+    help='A flag to show the command without running it.'
+)
+def rpict_command(
+        octree, view, rad_params, rad_params_locked, output, dry_run):
+    """Run rpict command for an input octree and a sensor grid.
```
devngc, line 43: Run rpict command for an input octree and a view file.

honeybee_radiance/cli/rpict.py
```diff
@@ -0,0 +1,74 @@
...
+
+    octree: Path to octree file.
+    sensor-grid: Path to sensor grid file.
```
devngc, line 46: view: path to the view file.

mostaphaRoudsari:
Hi @devngc, GitHub re-request doesn't work for some reason. I think I addressed all your comments here. Thank you for the review.

honeybee_radiance/cli/rpict.py
```diff
@@ -0,0 +1,74 @@
...
+@click.option(
+    '--dry-run', is_flag=True, default=False, show_default=True,
+    help='A flag to show the command without running it.'
+)
+def rpict_command(
+        octree, view, rad_params, rad_params_locked, output, dry_run):
+    """Run rpict command for an input octree and a view file.
+
+    octree: Path to octree file.
+    sensor-grid: Path to view file.
```
chriswmackey, line 46: I found this doesn't look as nice in the `--help` and sphinx as:
```
Run rpict command for an input octree and a view file.

\b
Args:
    octree: Path to octree file.
    sensor-grid: Path to view file.
```

honeybee_radiance/cli/rpict.py
```diff
@@ -0,0 +1,74 @@
...
+@click.option(
+    '--rad-params', show_default=True, help='Radiance parameters.'
+)
+@click.option(
+    '--rad-params-locked', show_default=True, help='Protected Radiance parameters. '
+    'These values will overwrite user input rad parameters.'
+)
+@click.option(
+    '--output', '-o', show_default=True, help='Path to output file. If a relative path'
+    ' is provided it should be relative to project folder.'
```
chriswmackey, line 35: You have `show_default` here but there's no default value. Also, we should agree on a standard for what we call this type of option since I have been calling this `--output-file` in at least 50 other commands [like this one](https://github.com/ladybug-tools/honeybee-energy/blob/master/honeybee_energy/cli/translate.py#L113-L115)

honeybee_radiance/cli/rpict.py
```diff
@@ -0,0 +1,74 @@
...
+
+@rpict.command('rpict')
+@click.argument(
+    'octree', type=click.Path(exists=True, file_okay=True, resolve_path=True)
+)
+@click.argument(
+    'view', type=click.Path(exists=True, file_okay=True, resolve_path=True)
+)
+@click.option(
+    '--rad-params', show_default=True, help='Radiance parameters.'
```
chriswmackey, line 27: Another `show_default` with no default value to show.

chriswmackey (commented):
Thanks @mostaphaRoudsari . Generally looking good and this helps me understand what you were saying earlier about how honeybee-radiance-command is used by honeybee-radiance.

honeybee_radiance/cli/rpict.py
mostaphaRoudsari: This is a tricky one - I'm all about making the libraries consistent. If we want to make this change we have to make it in other `honeybee-radiance` commands too like this one: https://github.com/ladybug-tools/honeybee-radiance/blob/c14f20e6ec3003e22a55d3667e9fe408f8290c01/honeybee_radiance/cli/raytrace.py#L33-L36

I think we might want to make the change from `--output` to `--output-file` all together in a separate commit. We have to update the plugins + recipes accordingly. But I'm all for it!

mostaphaRoudsari:
Hey @chriswmackey! Thank you for the comments. I addressed all except for one that I think should be addressed later. Let me know what you think.

chriswmackey (approved):
Looks good, @mostaphaRoudsari .

Do you have a sense of how many `--output` options we would have to change if we wanted to make things consistent?  As I said, I have ~50 `--output-file` and maybe ~10 `--log-file` in my code depending on whether the output is a usable file string or just a log. Is it fair to say that every command in `honeybee-radiance` CLI here is a `--output`?

mostaphaRoudsari:
Honestly, I don't remember but it is likely that every honeybee-radiance command or any other library that I have added a CLI for uses output and not output-file.

PR #516, opened 2021-04-06
docs(sphinx-conf): Fix CLI documentation process

santiagogaray:
- correctly handles commands without groups
- Check for group names inside cli\__init__.py

Hi @chriswmackey, updating to latest con.py as requested!


chriswmackey (approved):
LGTM!

PR #551, opened 2021-05-12
fix(sky): Add methods to serialize skies from strings

chriswmackey:
These can be used within all sky-based recipes so that we don't need a different recipe for each sky type. The new `honeybee-radiance sky from-string` command provides a single function to process any sky string into a sky file for recipes.

@mostaphaRoudsari , if you have any comments on the string representations, please let me know and I am happy to adjust them. I might merge this tomorrow so that [I can begin addressing this other schema issue](https://github.com/ladybug-tools/honeybee-schema/issues/204).

CC: @devngc

### DIFF
--- a/honeybee_radiance/cli/sky.py
+++ b/honeybee_radiance/cli/sky.py
@@ -5,11 +5,12 @@ import os
 import logging
 import json
 
-import honeybee_radiance.lightsource.sky as hbsky
-
-from honeybee_radiance.config import folders
+from ladybug.dt import DateTime
 from honeybee_radiance_command.gendaymtx import Gendaymtx, GendaymtxOptions
 from honeybee_radiance_command._command_util import run_command
+
+import honeybee_radiance.lightsource.sky as hbsky
+from honeybee_radiance.config import folders
 
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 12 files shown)


mostaphaRoudsari (commented):
Thanks @chriswmackey! I only have a comment on the styling of the string input. I think we can make it more similar to a command line input to gain some flexibility in how one can provide it - and how we can use it in Queenbee functions.

`CIE, 21 Jun 12:00, lat:41.78, lon:-87.75, type:0` -> `CIE 21 Jun 12:00 --lat 41.78 --lon -87.75 --type 0`

With this approach we can also path this definition to command line directly and take advantage of click's validation when loading the info. `honeybee-radiance sky CIE 21 Jun 12:00 --lat 41.78 --lon -87.75 --type 0` can be a valid command and the format follows what a command line is expected to look like. 

We can also breakdown the `sky-string` input. I personally think CIE vs ClimateBased should be sub-command and the rest should be arguments and options for each command. It will result in more commands but as I said it makes it closer to what a CLI command is expected to be.

This is how I would change the `from-string` command.

```python
@sky.command('CIE')
@click.argument('datetime', type=str, default='21 Jun 12:00')
@click.option('--lat', default=0)
@click.option('--lon', default=0)
@click.option('--tz', default=0)
@click.option('--type', default=0)
@click.option('--folder', help='Output folder.', default='.', show_default=True)
@click.option('--name', help='Sky file name.', default=None, show_default=True)
def cie_sky(datetime, lat, lon, tz, type, folder, name):
```

chriswmackey:
Thanks @mostaphaRoudsari ,

So I am perfectly fine with changing the style of the sky strings to be more command-like and getting the command-like string to match the command input of distinct `CIE`, `ClimateBased` and `CertainIrradiance' commands. However, I REALLY don't want to remove the `from-string` command because this kills the ability to write recipes that can accept any type of sky as input. Practically speaking, it means the we would have to offer at least 3 different recipes of "Point-in-time Grid Based" and 3 different recipes for "Point-in-time View Based" (6 in total) just to be able to support the three main types of skies.

Things just become much more streamlined from both the Grasshopper plugin side of things and the Pollination side of things when we have 2 versatile recipes rather than 6 semi-duplicated ones. This isn't to say that we couldn't use a command/ function that's specific to CIE or ClimateBased skies when someone is writing a recipe that is only intended to be used with one type of sky. But I just feel that we need something like the `from-string` method if we want to save ourselves from a lot of copy/paste work and the maintenance headache that comes with it. Do you agree?

chriswmackey:
I finished addressing all comments and I agree with @mostaphaRoudsari that the command structure is cleaner. Now, there's a 1-to-1 mapping between the string representations of the skies and the commands that can generate them. This also meant that I could make use of native Python `argparse` to write the methods that re-serialize skies from strings.

This is also all a lot more cleaner and intuitive than the JSON representations of the skies so I feel better about pointing people to this string schema and we'll just retire the current HBJSON documentation of Radiance sky assets.

PR #575, opened 2021-06-01
fix(postprocess): Allow occupied hours to be sun-down

chriswmackey:
This is a pretty big bug that I found in the postprocess module where we were computing the percent of sun-up occupied hours that met the illuminance criteria rather than the percent of all occupied hours that met the criteria. So an occupancy schedule with all night shift hours would have returned a DA of 100% instead of 0%. More info here:

https://discourse.ladybug.tools/t/annual-daylight-issue-with-1-2-2-version/14706

This one looks serious enough that we will have to update the recipe on Pollination.

CC: @mostapharoudsari

### DIFF
--- a/honeybee_radiance/postprocess/annualdaylight.py
+++ b/honeybee_radiance/postprocess/annualdaylight.py
@@ -54,19 +54,19 @@ def filter_schedule_by_hours(sun_up_hours, schedule=None):
         schedule: A list of 8760 values for the occupancy schedule.
 
     Returns:
-        List - A filtered version of the annual schedule that only includes the
-            sun-up-hours.
+        A tuple with two values.
 
+        occ_pattern -- A filtered version of the annual schedule that only
+            includes the sun-up-hours.
+        
+        total_hours -- A number for the total occupied hours of the schedule.
     """
     schedule = schedule or generate_default_schedule()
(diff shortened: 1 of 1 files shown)


honeybee_radiance/postprocess/annualdaylight.py
```diff
@@ -54,19 +54,19 @@ def filter_schedule_by_hours(sun_up_hours, schedule=None):
...
+        A tuple with two values.
 
+        occ_pattern -- A filtered version of the annual schedule that only
+            includes the sun-up-hours.
+        
+        total_hours -- A number for the total occupied hours of the schedule.
     """
     schedule = schedule or generate_default_schedule()
     occ_pattern = [schedule[int(h)] for h in sun_up_hours]
+    return occ_pattern, sum(schedule)
```
mostaphaRoudsari, line 66: Hi @chriswmackey, thank you for this fix. One comment here about using `sum(schedule)`. From the other lines in the code when you do `if is_occ < 0.5:` it looks like you are allowing non 0-1 values for occupancy. I don't think that's a good idea for daylight schedules but if we want to use them then we should calculate the sum based on the same logic: `sum(0 if is_occ < 0.5 else 0 for is_occ in schedule)`

```python
return occ_pattern, sum(0 if is_occ < 0.5 for is_occ in schedule else 1)
```

PR #618, opened 2021-07-09
style(metal): added a comment

devngc:


devngc:
This PR was done to show Mikkel how we send PRs from forks

PR #625, opened 2021-07-21
feat(dynamic): Unique glow modifier per aperture group

mikkelkp:
A unique modifier is needed for each aperture group when calculating multiple view matrices is one go. This replaces the default "white_glow" with a glow modifier with a unique name for the given aperture group.

### DIFF
--- a/honeybee_radiance/dynamic/state.py
+++ b/honeybee_radiance/dynamic/state.py
@@ -579,8 +579,7 @@ class RadianceSubFaceState(_RadianceState):
         The resulting string lacks modifiers and only includes the vmtx_geometry.
 
         Args:
-            modifier: A modifier (only identifier) assigned to the vmtx_geometry.
-                Default: white_glow.
+            modifier: A modifier object assigned to the vmtx_geometry. Default: white_glow.
             minimal: Boolean to note whether the radiance string should be written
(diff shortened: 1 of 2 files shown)


honeybee_radiance/dynamic/state.py
```diff
@@ -573,18 +573,20 @@ def scale(self, factor, origin=None):
...
 
-    def vmtx_to_radiance(self, minimal=False):
+    def vmtx_to_radiance(self, modifier=white_glow, minimal=False):
         """Generate a RAD string representation of this state's vmtx.
 
         The resulting string lacks modifiers and only includes the vmtx_geometry.
 
         Args:
+            modifier: A modifier (only identifier) assigned to the vmtx_geometry.
+                Default: white_glow.
```
chriswmackey, line 583: It's unclear what (only identifier) means here. Clearly you want the modifier object and not the text of the identifier, right?

chriswmackey (commented):
Generally looks ok to me. If you want to address the one comment and then I'll merge it when I know the changes will make it through our CI system (sorry it's under heavy construction now).

honeybee_radiance/dynamic/state.py
mikkelkp: That's correct. It should be fixed now.

PR #632, opened 2021-08-08
fix(state): add rfluxmtx header

mostaphaRoudsari:
This commit adds rfluxmtx header to state files. This header is requited for 3/5 phase studies.

The current implementation doesn't expose an overwrite from inside the group module. I _feel_ the chances that a user wants to overwrite the default values that are calculated from the geometry is minimal. That being said it might be a nice option to add for more advanced users in the future.

For this PR to work we need to merge this PR in first: https://github.com/ladybug-tools/honeybee-radiance-command/pull/193

@mikkelkp this basically does the same as what you proposed for adding the control parameters. I try to reuse the information from the geometry and the modifier of the aperture group.


honeybee_radiance/dynamic/group.py
```diff
@@ -289,6 +291,46 @@ def states_json_list(self):
...
+        Args:
+            sampling: Set hemisphere sampling type. Acceptable inputs for hemisphere
+                sampling type are:
+                * u for uniform.(Usually applicable for ground).
+                * kf for klems full.
+                * kh for klems half.
+                * kq for klems quarter.
+                * rN for Reinhart - Tregenza type skies. N stands for subdivisions and
+                defaults to 1.
+                * scN for shirley-chiu subdivisions.
```
chriswmackey, line 309: You need line breaks around this bulleted list for it to render correctly in Sphinx.

honeybee_radiance/dynamic/group.py
```diff
@@ -289,6 +291,46 @@ def states_json_list(self):
...
+        if up_direction is None:
+            up_direction = state.vmtx_geometry.plane.y
+        if sampling is None:
+            try:
+                sampling = state.modifier.sampling_type
+            except AttributeError:
+                raise ValueError(
+                    'Rfluxmtx control parameters can only be generated for states '
+                    'with BSDF modifier. Current modifier: %s' % state.modifier
+                )
```
chriswmackey, line 330: Does this mean that we only support window groups with BSDF modifiers? If so, then this probably merits a check when we assign a `group_identifier` to an aperture to make sure that the Aperture has a BSDF assigned to it. It's better for users to get the error when they are creating the window group rather than getting all of the way to running the recipe and having it fail with a denser luigi/argo message.

Also, you should probably do `type(state.modifier)` or `state.modifier.__class__.__name__` since the repr of Modifiers isn't going to be that clear in an error message like this. It's probably also helpful for you to include the identifier of the modifier. Like:

```
Apertures in Window Groups must have a BSDF modifier. Current modifier "{}" is a {}.'.format(self.modifier, self.modifier.__class__.__name__)
```

honeybee_radiance/modifier/material/bsdf.py
```diff
@@ -233,6 +233,30 @@ def angle_basis(self, value):
...
+            ValueError: If angle basis is ``TensorTree``.
+
+        """
+        _mapper = {
+        'Klems Full': 'kf', 'Klems Half': 'kh', 'Klems Quarter': 'kq'
+        }
+        if self.angle_basis == 'TensorTree':
+            raise ValueError(
+                'There is no valid sampling type for Tensor Tree BSDFs. they can only '
+                'be used for a in-scene modifiers.'
```
chriswmackey, line 255: It seems a little odd to get an exception raised just for checking a property. Maybe we should just return `None` when checking the property and then, if this is illegal for a given workflow, we raise the exception in that workflow.

chriswmackey (commented):
A few suggestions.

honeybee_radiance/dynamic/group.py
mostaphaRoudsari: Thank you @chriswmackey! I made the adjustment to the message to be more clear.

> Does this mean that we only support window groups with BSDF modifiers?

The short answer is no. We support aperture groups with non-BSDF materials but in a 2-Phase recipe and not a 3 or 5 Phase which is when this error will be raised.

I agree that we should give an error sooner than later. Since this error is recipe-specific, the best place to raise an error is probably the model handler of the recipes. We can run a quick check on the model and raise an error before creating the run.

PR #655, opened 2021-10-03
feat(cli grid): add two new commands to split and merge grid folders

mostaphaRoudsari:
This change will cause an error until we merge this one:

https://github.com/ladybug-tools/honeybee-radiance-folder/pull/109

### DIFF
--- a/honeybee_radiance/cli/grid.py
+++ b/honeybee_radiance/cli/grid.py
@@ -9,6 +9,8 @@ import json
 from honeybee.model import Model
 from honeybee.units import parse_distance_string
 import honeybee_radiance.sensorgrid as sensorgrid
+from honeybee_radiance_folder.gridutil import redistribute_sensors, \
+    restore_original_distribution
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


honeybee_radiance/cli/grid.py
```diff
@@ -186,3 +188,93 @@ def from_rooms(model_json, grid_size, offset, include_mesh, keep_out, room,
...
+@click.argument(
+    'output-folder',
+    type=click.Path(file_okay=False, dir_okay=True, resolve_path=True))
+@click.argument('grid-count', type=int)
+@click.option(
+    '--min-sensor-count', '-msc', help='Minimum number of sensors in each output grid. '
+    'Use this number to ensure the number of sensors in output grids never gets very '
+    'small. To ignore this limitation set the value to 1. Default: 2000.', type=int,
+    default=2000)
+def distribute_sensor_grids(input_folder, output_folder, grid_count, min_sensor_count):
```
chriswmackey, line 206: The few cases where I didn't name the function with the same name as the command haven't turned out so well for me over time. Granted, I know it's just a matter of style but having different names between commands and functions has been a little like naming a hidden property on a class something different than it's public property. It's not that it's inherently wrong. I just wish in hindsight that I did it differently since it make it tougher to know which command is being run in each of the test among other things.

chriswmackey (commented):
I agree that `split-folder` and `merge-folder` are good names for the commands and they make sense in the context of the command group.

PR #663, opened 2021-10-27
Add new command for view matrix calculation

mostaphaRoudsari:
@ladybug-tools/radiance-developers - here is the command that I wrote during the call. I cleaned up the text and added a test for the command.


honeybee_radiance/cli/multiphase.py
```diff
@@ -0,0 +1,113 @@
...
+        options = RfluxmtxOptions()
+        # parse input radiance parameters
+        if rad_params:
+            options.update_from_string(rad_params.strip())
+        # overwrite input values with protected ones
+        if rad_params_locked:
+            options.update_from_string(rad_params_locked.strip())
+
+        if not sensor_count:
+            raise ValueError("for the time-being sensor count must be provided.")
```
chriswmackey, line 88: I'm inclined to think that we shouldn't merge this until we address this. If the sensor-count is not provided, we just need just a few lines of code to count the number of lines in the file, right?

I think we should probably also add something to the input description to say why you would supply this sensor-count. It's just to make the calculation faster, correct?

chriswmackey (commented):
I just have one comment. Otherwise it's all good.

honeybee_radiance/cli/multiphase.py
mostaphaRoudsari: Thanks, @chriswmackey!

@mikkelkp, is it something that you can address? It will be a helper function. We already have such functions that read sensor grids from a file.

If that's the case my suggestion is to merge this one and then update this command and a couple of other commands which have the same error message.

mikkelkp: Sure, @mostaphaRoudsari. I assume you think of [from_file](https://github.com/ladybug-tools/honeybee-radiance/blob/8804c79a991c3160484b510198de14a59bda028e/honeybee_radiance/sensorgrid.py#L182) and then get the sensor count by the count property.

I can add it to the other commands as well.

mostaphaRoudsari: Hi @mikkelkp! Yes. But with the difference that you don't need to create the sensor grids which is slower. You just need to go through the file and count the number of lines while making sure you don't count the comment lines if any.

mostaphaRoudsari:
I merged this one in so we can merge the rest of the features that are built on top of this one.

PR #666, opened 2021-11-05
feat(view): Allow 360 fisheyes for view type -vta

mikkelkp:
I put this in the `standardize_fisheye` method to avoid editing the HB View component in GH.


chriswmackey (approved):
LGTM! Sorry I didn't see this earlier. We should merge this.

PR #667, opened 2021-11-10
feat(view): add option for appending view to header

mikkelkp:


honeybee_radiance/cli/view.py
```diff
@@ -139,6 +139,10 @@ def split_view(view, count, resolution, skip_overture, octree, rad_params,
     file_okay=False, dir_okay=True, resolve_path=True))
 @click.argument('base-name', type=str)
 @click.argument('extension', default='.unf', type=str)
+@click.option('--view', '-vf', type=click.Path(
+        exists=True, file_okay=True, dir_okay=False, resolve_path=True),
+        help='Full path to input sensor view file.'
```
mostaphaRoudsari, line 144:
```suggestion
        help='Full path to the original view file.'
```

honeybee_radiance/cli/view.py
```diff
@@ -191,7 +195,12 @@ def merge_view(input_folder, base_name, extension, scale_factor, folder, name):
 
         # search for a single .vf in the folder and, if it's found, grab the info
         views = sorted(f for f in os.listdir(input_folder) if f.endswith('.vf'))
-        if len(views) == 1:  # replace the header with the info in the view
+        if view:
+            view_obj = View.from_file(view)
+            getinfo = Getinfo(output=output_file)
+            getinfo.options.a = 'VIEW= {}'.format(view_obj)
+            pfilt.pipe_to = getinfo
```
mostaphaRoudsari, line 202: Really nice! I imagine you have tested this already but It will be good to add a test for this functionality.

honeybee_radiance/cli/view.py
```diff
@@ -191,7 +195,12 @@ def merge_view(input_folder, base_name, extension, scale_factor, folder, name):
 
         # search for a single .vf in the folder and, if it's found, grab the info
         views = sorted(f for f in os.listdir(input_folder) if f.endswith('.vf'))
-        if len(views) == 1:  # replace the header with the info in the view
+        if view:
```
mostaphaRoudsari, line 198: We probably only want to do this when there are more than one view.

```suggestion
        if view and len(views) != 1:
```

honeybee_radiance/cli/view.py
```diff
@@ -191,7 +195,12 @@ def merge_view(input_folder, base_name, extension, scale_factor, folder, name):
...
+            view_obj = View.from_file(view)
+            getinfo = Getinfo(output=output_file)
+            getinfo.options.a = 'VIEW= {}'.format(view_obj)
+            pfilt.pipe_to = getinfo
+        elif len(views) == 1:  # replace the header with the info in the view
             view_obj = View.from_file(os.path.join(input_folder, views[0]))
             getinfo = Getinfo(output=output_file)
             getinfo.options.a = 'VIEW= {}'.format(view_obj)
```
mostaphaRoudsari, line 206: What if there are multiple views and the original view is not provided. Can we do the quick math as suggested by @chriswmackey to re-create the original view?

mostaphaRoudsari (commented):
Hi @mikkelkp! Looks great. See my comments.

honeybee_radiance/cli/view.py
mikkelkp: Yes, I ran the command with the new view option and it appended correctly to the header.

honeybee_radiance/cli/view.py
mikkelkp: I believe if there are multiple views, we will assume that those are subviews. Then we can do the math, basically the reverse of the [grid](https://github.com/ladybug-tools/honeybee-radiance/blob/0de8d9a5ee99332b9046c133e488e9846f0f3ba1/honeybee_radiance/view.py#L632) method that I also mentioned.
If there are multiple views and they are **not** subviews, e.g. different view point or directions, then the reverse grid method will fail.

PR #686, opened 2021-12-16
fix(writer): Discount AirBoundaries from Radiance

chriswmackey:
It seems that, even with the 100% transparent modifier, the AirBoundaries still manage to show up in the results (perhaps because there are two of them for interior Faces?) In any even, discounting them definitely gets the results to look as expected.


chriswmackey:
I realized that this issue was due to something else and this change is not necessary. I'm just going to close the PR. Things are fine with the current Trans air_boundary modifier.

PR #699, opened 2022-01-21
feat(multiphase): add command for daylight matrix calculation

mikkelkp:


honeybee_radiance/cli/multiphase.py
```diff
@@ -112,3 +112,87 @@ def view_matrix_command(
         sys.exit(1)
     else:
         sys.exit(0)
+
+
+@multi_phase.command("daylight-matrix")
```
mostaphaRoudsari, line 117: Should we call this daylight coefficient instead? I think that's technically what we are calculating here - or even view coefficient. It's not really daylight coefficient depending on how it is being used.

What do you think @built-tools?

honeybee_radiance/cli/multiphase.py
```diff
@@ -112,3 +112,87 @@ def view_matrix_command(
...
+    output,
+    dry_run,
+):
+    """Calculate daylight matrix for a sender file.
+
+    \b
+    Args:
+        sender_file: Path to sender file.
+        receiver_file: Path to receiver file.
+        octree: Path to octree file.
```
mostaphaRoudsari, line 163: The sender and receiver file can use a little bit more information in the description to clarify how they look like or what can they be.

tests/cli/multiphase_test.py
```diff
@@ -28,3 +28,22 @@ def test_view_matrix_command():
...
+    output = './tests/assets/temp/east_window_classroom.dmx'
+    nukedir(output_folder)
+    cmd_args = [
+        sender_file, receiver_file, input_oct,
+        '--rad-params', '-ab 3 -ad 1000 -lw 1e-05 -c 1000',
+        '--output', output
+    ]
+    result = runner.invoke(daylight_matrix_command, cmd_args)
+    assert result.exit_code == 0
+    assert os.path.isfile(output)
```
mostaphaRoudsari, line 48: We probably want to add another check to make sure the file size is larger than an empty file with an error.

tests/cli/multiphase_test.py
```diff
@@ -28,3 +28,22 @@ def test_view_matrix_command():
...
+    runner = CliRunner()
+    sender_file = './tests/assets/multi_phase/east_window_classroom..sender.rad'
+    receiver_file = './tests/assets/multi_phase/rflux_sky.sky'
+    input_oct = './tests/assets/multi_phase/scene.oct'
+    output_folder = './tests/assets/temp'
+    output = './tests/assets/temp/east_window_classroom.dmx'
+    nukedir(output_folder)
+    cmd_args = [
+        sender_file, receiver_file, input_oct,
+        '--rad-params', '-ab 3 -ad 1000 -lw 1e-05 -c 1000',
```
mostaphaRoudsari, line 43: -ab 3 might be too much for a test. I don't know how long it takes to run this command but it shouldn't take more than a few seconds for a test if you ask me.

mostaphaRoudsari (commented):
Looks clean! I'm a bit out of the zone of Radiance commands for being trusted to catch all the possible issues/edge cases but from what I can see the overall approach makes sense.

I left a comment about the naming of the command. If @built-tools and you agree on this name then I'm fine with it but it looked like the name can be more generic.

built-tools:
@mikkelkp and @mostaphaRoudsari , I don't see any harm in merging this. However, I think that **strictly from a user perspective**, all the daylight recipes that eventually get developed would benefit from having access to matrix creation processes as functions rather than cli commands. 

It seems unlikely (to me) that someone interacting  would run a honeybee/pollination cli command to just create a daylight matrix or a view matrix. It appears more logical that they would be interested in running either a complete n-phase process or solving an entire model folder. 

Anyway, let's continue this discussion on Basecamp. I will create a new discussion for this in a day or two.

honeybee_radiance/cli/multiphase.py
mikkelkp: @mostaphaRoudsari, you are right that the name can be more generic as the command is essentially the most simple use of rfluxmtx. As such I changed the name to 'flux-transfer'.

mostaphaRoudsari:
Thank you @mikkelkp! I'm merging this in.

PR #706, opened 2022-02-17
feat(aBSDF): Add aBSDF modifier to library

mikkelkp:


mostaphaRoudsari (commented):
Hi @mikkelkp, from what I remember this looks almost identical to the BSDF material. Is there a reason that we don't subclass this class from BSDF?

mikkelkp:
Hi @mostaphaRoudsari, the only difference is that aBSDF has no thickness. I tried to subclass from BSDF in a new commit.

honeybee_radiance/modifier/material/absdf.py
```diff
@@ -0,0 +1,261 @@
...
+
+    # TODO(): compress file content: https://stackoverflow.com/a/15529390/4394669
+    def __init__(self, bsdf_file, identifier=None, up_orientation=None, modifier=None, 
+                 function_file='.', transform=None, angle_basis=None, dependencies=None):
+        """Create aBSDF material."""
+        # thickness is ignored in aBSDF class
+        thickness = None
+        BSDF.__init__(self, bsdf_file, identifier, up_orientation, thickness,
+                 modifier, function_file, transform, angle_basis,
+                 dependencies)
```
mostaphaRoudsari, line 83: I see. The right way of doing this is probably to use aBSDF as the base class and subclass BSDF from it. In that scenario, you can add thickness as the extra property to aBSDF. I would say we should do that if it is not too much work.

mostaphaRoudsari (commented):
Hi @mikkelkp! This is good but technically we should implement it the other way around.

mikkelkp:
@mostaphaRoudsari, I subclassed BSDF from aBSDF. I think I got it right, at least I am able to create both modifiers without any errors.

honeybee_radiance/modifier/material/absdf.py
```diff
@@ -0,0 +1,468 @@
...
+from .materialbase import Material
+import honeybee.typing as typing
+from honeybee.config import folders
+import ladybug_geometry.geometry3d.pointvector as pv
+
+
+class aBSDF(Material):
+    """Radiance aBSDF material.
+
+    .. code-block:: shell
```
mostaphaRoudsari, line 17: Can we add a small  note here about the aBSDF material and how it is different from BSDF - the only document that I know of is Greg's presentation during the Radiance workshop.

honeybee_radiance/modifier/material/bsdf.py
```diff
@@ -1,17 +1,15 @@
 """Radiance BSDF Material.
 
-http://radsite.lbl.gov/radiance/refer/ray.html#Glass
+https://floyd.lbl.gov/radiance/refer/ray.html#BSDF
 """
 from __future__ import division
 
 import os
-from .materialbase import Material
+from honeybee_radiance.modifier.material.absdf import aBSDF
```
mostaphaRoudsari, line 8: You should be able to use relative import here.

```suggestion
from .absdf import aBSDF
```

honeybee_radiance/modifier/material/bsdf.py
```diff
@@ -74,31 +72,20 @@ class BSDF(Material):
         * is_modifier
         * is_material
     """
-    __slots__ = ('_bsdf_file', '_up_orientation', '_thickness', '_function_file',
-                 '_transform', '_angle_basis', '_front_diffuse_reflectance',
-                 '_back_diffuse_reflectance', '_diffuse_transmittance')
+
+    __slots__ = ('_thickness')
```
mostaphaRoudsari, line 76: A tuple with a single item and no ',' will be considered a string!

```suggestion
    __slots__ = ('_thickness',)
```

honeybee_radiance/modifier/material/bsdf.py
```diff
@@ -74,31 +72,20 @@ class BSDF(Material):
...
         """Create BSDF material."""
-        identifier = identifier or '.'.join(os.path.split(bsdf_file)[-1].split('.')[:-1])
-
-        Material.__init__(self, identifier, modifier=modifier,
-                          dependencies=dependencies)
-
-        self.bsdf_file = bsdf_file
-        self.angle_basis = angle_basis
-        self.up_orientation = up_orientation
         self.thickness = thickness or 0
```
mostaphaRoudsari, line 83: You should probably move this line after you initiate the class in line 84.

honeybee_radiance/modifier/material/bsdf.py
```diff
@@ -74,31 +72,20 @@ class BSDF(Material):
...
-        self._front_diffuse_reflectance = None
-        self._back_diffuse_reflectance = None
-        self._diffuse_transmittance = None
-        self._update_values()
+        aBSDF.__init__(self, bsdf_file, identifier, up_orientation, modifier, 
+                 function_file, transform, angle_basis, dependencies)
 
+        self._update_values()
+        
     def _update_values(self):
```
mostaphaRoudsari, line 89: Do we need this method here? If we don't make any changes to it then we can simply use the one from aBSDF class.

mostaphaRoudsari (commented):
Hi @mikkelkp, it looks good to me. I had a few minor comments.

Also can we at least add one test to make sure everything works as expected before we merge this in?

mostaphaRoudsari:
@mikkelkp, what is the status of this PR? I think we should finalize the changes and merge them in sooner than later.

honeybee_radiance/modifier/material/bsdf.py
mikkelkp: Hi @mostaphaRoudsari, I didn't see you added all these comments. I corrected your suggestions except this one, as it will run `_update_values()` when initiating the class, but the thickness attribute is not yet set, which is needed in line 92 in `_update_values()`.

PR #708, opened 2022-03-01
feat(dcglare): Add dcglare cli command for two-phase

mikkelkp:


tests/cli/threephasemtx_test.py
```diff
@@ -1,4 +1,4 @@
-"""Test cli translate module."""
+"""Test cli threephase module."""
```
mostaphaRoudsari, line 1: Thank you for fixing this!

honeybee_radiance/cli/glare.py
```diff
@@ -0,0 +1,104 @@
...
+@click.option(
+    '--threshold-factor', '-b', type=float,
+    help='If larger than 100, it is used as constant threshold in cd/m2. If less than '
+    'or equal to 100, this factor multiplied by the average luminance in each view will '
+    'be used as threshold for detecting the glare sources (not recommended). The '
+    'default value is 2000 (fixed threshold method).'
+)
+@click.option(
+    '--occupancy-schedule', '-sf', type=click.Path(exists=True, dir_okay=False, 
+    resolve_path=True),
```
mostaphaRoudsari, line 45: In order to use this in a recipe and keep it optional, you have to relax the `exists=True` input so you can deal with it inside the command itself.

mostaphaRoudsari (approved):
Hi @mikkelkp, Look great. I had a minor comment which will mainly be important when you develop a Pollination Function using the command.

Feel free to to edit and merge this in.

PR #709, opened 2022-03-01
fix(octree): Add an option to keep aperture groups as black

chriswmackey:
I feel like blacked-out apertures is a better default behavior for all of the recipes that aren't capable of modeling dynamic apertures. At least this will make it clear to people when a certain recipe does not support dynamic groups.

It's also the default that I need in the comfort mapping workflows.

### DIFF
--- a/honeybee_radiance/cli/octree.py
+++ b/honeybee_radiance/cli/octree.py
@@ -24,13 +24,19 @@ def octree():
     ' is provided it should be relative to project folder.'
 )
 @click.option(
-    '--default/--black', is_flag=True, default=True, show_default=True,
-    help='Flag to note wheather the octree should be created with black materials.'
+    '--default/--black',  ' /-b', default=True, show_default=True,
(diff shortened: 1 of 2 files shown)


honeybee_radiance/cli/octree.py
```diff
@@ -58,8 +66,14 @@ def create_octree_from_folder(
...
                 scene_files += aperture_files
-            except FileNotFoundError:
+            except Exception:
                 pass  # no apertures available in the model
+        if black_groups:
+            try:
+                group_files = model_folder.aperture_group_files_black()
+                scene_files += group_files
+            except Exception:
+                pass  # no aperture groups available in the model
```
mostaphaRoudsari, line 76: Hey @chriswmackey, Having a generic Exception/pass is a really dangerous pattern. Why do you need to do this? Is there a better and more explicit way to handle this? There are papers and articles that say why you should never do such a thing.

PR #715, opened 2022-03-14
feat(threephase): Allow rgb to illuminance conversion in cli command

mikkelkp:
Should only be merged once https://github.com/ladybug-tools/honeybee-radiance-command/pull/236 has been merged. The requirements should be updated once there is a new release of `honeybee-radiance-command`.

For now the test I added fails because it relies on the above changes to `honeybee-radiance-command`.


mostaphaRoudsari (approved):
LGTM! Feel free to merge it after updating the dependencies. The new version for honeybee-radiance-command will be 1.22.0

PR #719, opened 2022-03-18
feat(multiphase): Add command for combined octrees and grids

mikkelkp:
Needs `honeybee-radiance-folder==2.10.4` or whatever version this PR is released under: https://github.com/ladybug-tools/honeybee-radiance-folder/pull/130

### DIFF
--- a/honeybee_radiance/cli/multiphase.py
+++ b/honeybee_radiance/cli/multiphase.py
@@ -19,6 +19,7 @@ from honeybee_radiance_folder import ModelFolder
 from honeybee_radiance_folder.gridutil import redistribute_sensors
 from honeybee.aperture import Aperture
 from honeybee_radiance.geometry.polygon import Polygon
+from .octree import _generate_octrees_info
 
 from .threephase import three_phase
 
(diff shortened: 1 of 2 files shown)


honeybee_radiance/cli/multiphase.py
```diff
@@ -505,3 +507,179 @@ def _aperture_view_factor(project_folder, apertures, size=0.2, ambient_division=
...
+            'To generated octrees for a 5 Phase study you must provide a sunpath.'
+        )
+
+    phases = {
+        '2': ['two_phase'],
+        '3': ['two_phase', 'three_phase'],
+        '5': ['two_phase', 'three_phase', 'five_phase']
+    }
+
+    try:
```
mostaphaRoudsari, line 581: This is becoming a really long CLI command.

What we usually do is to create separate smaller functions for each step so we can test them separately and then we use them in the CLI command. Can you think of a logical way to break this down and write a few tests for each part?

mostaphaRoudsari (commented):
Thank you @mikkelkp! See my comment.

PR #725, opened 2022-03-23
fix(dc): Add coefficient command without sky mtx

mikkelkp:
This is needed for imageless glare.

### DIFF
--- a/honeybee_radiance/cli/dc.py
+++ b/honeybee_radiance/cli/dc.py
@@ -361,10 +361,8 @@ def rfluxmtx_command_without_postprocess(
 
         options.update_from_string('-aa 0.0 -y {}'.format(sensor_count))
 
-        cmd = Rfluxmtx(options=options, octree=octree, sensors=sensor_grid, 
-                       receivers=sky_dome)
-        if output:
-            cmd.output = output
+        cmd = Rfluxmtx(options=options, output=output, octree=octree, 
+                       sensors=sensor_grid, receivers=sky_dome)
 
         if dry_run:
             click.echo(cmd)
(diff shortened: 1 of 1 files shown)


honeybee_radiance/cli/dc.py
```diff
@@ -242,10 +242,10 @@ def rfluxmtx_command_with_postprocess(
 
         options = RfluxmtxOptions()
         # parse input radiance parameters
-        if rad_params.strip():
+        if rad_params:
```
mostaphaRoudsari, line 245: Checking with `.strip()` makes sure we handle cases like this one `     `. Is it a reason that you decided to remove them?

honeybee_radiance/cli/dc.py
```diff
@@ -287,3 +287,95 @@ def rfluxmtx_command_with_postprocess(
...
+)
+@click.option(
+    '--keep-header/--remove-header', ' /-h-', is_flag=True, default=True,
+    help='A flag to keep or remove the header from the output file.'
+)
+@click.option(
+    '--dry-run', is_flag=True, default=False, show_default=True,
+    help='A flag to show the command without running it.'
+)
+def rfluxmtx_command_without_postprocess(
```
mostaphaRoudsari, line 331: Is it possible to add an optional flag to our existing rfluxmtx command, assuming we have one, instead of creating a new command? `--postprocess/--no-postprocess`?

mostaphaRoudsari (commented):
LGTM! I have two comments about the development decisions.

honeybee_radiance/cli/dc.py
mikkelkp: If I don't use the option '--rad-params' in command line it fails. Without `.strip()` it works even with '--rad-params ""' and when I exclude the option completely.

image

mostaphaRoudsari: Good catch!

honeybee_radiance/cli/dc.py
mikkelkp: I see the name may be a bit misleading. The existing command is with a sky matrix (argument) and optional postprocess. I don't need the sky matrix or any postprocessing.

PR #735, opened 2022-04-06
fix(postprocess): Add glare autonomy commands and functions

mikkelkp:


mostaphaRoudsari (commented):
Hi @mikkelkp, It looks good to me but I _feel_ we should create a new module for all the post-process that is related to annual glare studies. They don't really belong under the `annualdaylight` module.

mostaphaRoudsari (approved):
LGTM! Thanks.

PR #737, opened 2022-04-12
fix(multiphase): Add exclude static option

mikkelkp:
I also renamed the command from octrees-grids to prepare-dynamic


honeybee_radiance/cli/multiphase.py
```diff
@@ -510,7 +509,7 @@ def _aperture_view_factor(project_folder, apertures, size=0.2, ambient_division=
         sys.exit(0)
 
 
-@multi_phase.command('octrees-grids')
+@multi_phase.command('prepare-dynamic')
```
mostaphaRoudsari, line 512: I like the new name better but this is a breaking change. Am I right that we are currently only using this command in the multiphase recipes? If that's the case, then I'm fine with it, otherwise, we should probably don't change the name as it can break other recipes that are shipped under the lbt-recipes.

mostaphaRoudsari (approved):
Hi @mikkelkp! The changes look fine to me. I haven't seen how it changes the recipe but I like the idea of being able to remove the static. I don't know how we got it to work for 3 Phase then. Did you have to change that too?

honeybee_radiance/cli/multiphase.py
mikkelkp: It is only used in three-phase, so it will not break anything.

mikkelkp:
> Hi @mikkelkp! The changes look fine to me. I haven't seen how it changes the recipe but I like the idea of being able to remove the static. I don't know how we got it to work for 3 Phase then. Did you have to change that too?

I made the changes to run three-phase with 'prepare-dynamic' command locally. I can update three-phase once this PR goes through.

PR #738, opened 2022-04-13
fix(annualglare): Add function to calculate GA from folder

mikkelkp:
The same way that annual daylight metrics can be calculated from the results folder.


chriswmackey (approved):
This looks good to me, @mikkelkp . I like you followed similar conventions to what we did for annual daylight post-processing.

Lets merge!

PR #741, opened 2022-04-20
fix(schedule): Add command to convert EPW to daylight hours

mikkelkp:
I am a bit conflicted about creating a new module 'schedule' just for one command, but I didn't know where else to put it.


honeybee_radiance/cli/schedule.py
```diff
@@ -0,0 +1,51 @@
...
+    pass
+
+
+@schedule.command('epw-to-daylight-hours')
+@click.argument(
+    'epw', type=click.Path(exists=True, dir_okay=False, resolve_path=True))
+@click.option(
+    '--schedule-name', '-sn', help='Output file name for schedule.',
+    type=click.STRING, default='daylight_hours', show_default=True
+)
```
mostaphaRoudsari, line 23: If you are doing it like this, then you probably want to give the user bother the folder and the file name. We can also call it `name` instead of `schedule-name` since there are no other names. Here is an example:

https://github.com/ladybug-tools/honeybee-radiance/blob/f328bfcc2906afa8eb2aadffcbb8f1f4476b060f/honeybee_radiance/cli/sky.py#L156-L157

honeybee_radiance/cli/schedule.py
```diff
@@ -0,0 +1,51 @@
...
+        _illuminance, indices = diffuse_horizontal_illuminance.highest_values(4380)
+        values = ['1' if i in indices else '0' for i in range(8760)]
+
+        # write csv
+        file_path = '%s.csv' % schedule_name
+        with open(file_path, 'w') as fp:
+            fp.write('\n'.join(values))
+
+    except Exception:
+        _logger.exception('Failed to generate daylight hours.')
```
mostaphaRoudsari, line 48:
```suggestion
        _logger.exception('Failed to generate daylight hours schedule.')
```

mostaphaRoudsari (approved):
Thanks, @mikkelkp! It looks good to me. Two minor comments. Feel free to merge it once you address those.

PR #742, opened 2022-04-21
fix(postprocess): Add EN 17037 calculations

mikkelkp:
Just a draft for now.


honeybee_radiance/cli/postprocess.py
```diff
@@ -393,6 +394,53 @@ def annual_metrics(
...
+    recommendation: minimum, medium, high.
+
+    \b
+    Args:
+        folder: Results folder. This folder is an output folder of annual
+            daylight recipe. Folder should include grids_info.json and sun-up-hours.txt.
+            The command uses the list in grids_info.json to find the result files for
+            each sensor grid.
+        schedule: An annual schedule for 8760 hours of the year as a list of values. This
+            should be a daylight hours schedule.
```
chriswmackey, line 431: This description is not correct. You are expecting a CSV file, right? Not a list.

honeybee_radiance/postprocess/en17037.py
```diff
@@ -0,0 +1,264 @@
...
+    for is_occ, value in zip(occ_pattern, values):
+        if is_occ == 0:
+            continue
+        if value > threshold:
+            da += 1
+
+    return _percentage(da, total_hours)
+
+
+def metrics_to_files(
```
chriswmackey, line 34: I know that you are trying to follow the example of annual-daylight but it might be a good idea to name this method with a slightly different name just to avoid one method overwriting another if they are both imported to the same place. Maybe we should call this `en17037_metrics_to_files`

honeybee_radiance/postprocess/en17037.py
```diff
@@ -0,0 +1,264 @@
+"""Functions for post-processing EN 17037 daylight outputs."""
+import json
+import os
+
+from .annual import filter_schedule_by_hours, _process_input_folder
+
+
+def _metrics(values, occ_pattern, threshold, total_hours):
```
chriswmackey, line 8: This name doesn't seem quite correct. There's only one metric, right? Maybe we should just call it `_daylight_autonomy`

honeybee_radiance/postprocess/en17037.py
```diff
@@ -0,0 +1,264 @@
...
+    """Calculate annual metrics for a sensor.
```
chriswmackey, line 9: And this description does not seem correct

honeybee_radiance/postprocess/en17037.py
```diff
@@ -0,0 +1,264 @@
...
+
+    Args:
+        values: Hourly illuminance values as numbers.
+        occ_pattern: A list of 0 and 1 values for hours of occupancy.
+        threshold: Threshold value for daylight autonomy.
+        total_occupied_hours: An integer for the total number of occupied hours,
```
chriswmackey, line 15: The name of this input is `total_hours`. Not `total_occupied_hours`

honeybee_radiance/postprocess/en17037.py
```diff
@@ -0,0 +1,264 @@
...
+            which can be used to avoid having to sum occ pattern each time.
+
+    Returns:
+        daylight autonomy
+    """
+    def _percentage(in_v, occ_hours):
+        return round(100.0 * in_v / occ_hours, 2)
```
chriswmackey, line 22: Given that this is only one line of code that we use once in this module, maybe we don't need a method for it.

chriswmackey (commented):
It generally looks good, @mikkelkp . I just have a few comments related to docs and naming.

PR #764, opened 2022-07-04
fix(static apertures): change the default name to __static_apertures__

mostaphaRoudsari:
This will ensure that they are always called the same name in the model, model folder and results folder.


mostaphaRoudsari:
cc: @mikkelkp, @chriswmackey

PR #778, opened 2022-10-09
fix(dynamic daylight): use full_id instead of identifier

mostaphaRoudsari:
Hi, @mikkelkp - I don't dare to merge this in as it might break other functionalities but we should use the `full_id` as the identifier for sensor grids instead of the `identifier`.

Otherwise, we will face issues like this: https://github.com/ladybug-tools/honeybee-radiance-postprocess/pull/67


mikkelkp (approved):
As far as I am aware it should not break anything.

mostaphaRoudsari:
I merged it in! I leave the checks on the post process library to you.

PR #785, opened 2022-10-24
fix(octree): Set scene files as empty list

mikkelkp:
This is needed for certain models to run in the cloud.

### DIFF
--- a/honeybee_radiance/cli/octree.py
+++ b/honeybee_radiance/cli/octree.py
@@ -73,7 +73,10 @@ def create_octree_from_folder(
     model_folder = ModelFolder.from_model_folder(folder)
     try:
         black_out = False if default else True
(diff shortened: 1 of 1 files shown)


honeybee_radiance/cli/octree.py
```diff
@@ -73,7 +73,10 @@ def create_octree_from_folder(
     model_folder = ModelFolder.from_model_folder(folder)
     try:
         black_out = False if default else True
-        scene_files = model_folder.scene_files(black_out=black_out)
+        try:
+            scene_files = model_folder.scene_files(black_out=black_out)
```
mostaphaRoudsari, line 77: @mikkelkp, we probably want to add this check to the honeybee-radiance-folder library, and fix the issue in the source.

PR #817, opened 2023-04-12
fix(writer): Remove air boundaries from Radiance folder

mikkelkp:


honeybee_radiance/writer.py
```diff
@@ -645,43 +647,64 @@ def _write_static_files(
         punched_verts: Boolean noting whether punched geometry should be written
         minimal: Boolean noting whether radiance strings should be written minimally.
     """
+    def is_air_boundary(face):
+        if isinstance(face, Face):
+            if isinstance(face.type, AirBoundary):
+                return True
+        else:
+            return False
```
chriswmackey, line 655: Just a suggestion. This might be a bit more succinct and Pythonic:
``
def is_air_boundary(face):
    return isinstance(face, Face) and isinstance(face.type, AirBoundary)
```
That will work for all cases given that it first checks that it's a Face before checking the type.

chriswmackey (commented):
LGTM!

PR #879, opened 2024-10-24
fix(viewfactor): Add ShadeMesh to scene file

mikkelkp:


chriswmackey (approved):
Thanks for finding (and fixing) this bug, @mikkel. This one might have stayed here for a while without both Jon and you looking into this.

Merged!

PR #889, opened 2025-02-26
fix(multiphase): Add option to create groups if none exists

mikkelkp:


honeybee_radiance/cli/multiphase.py
```diff
@@ -947,17 +952,29 @@ def add_aperture_group_blinds_command(
...
+
+        unique_aperture_groups = get_unique_aperture_groups(model)
+
+        if not unique_aperture_groups and create_groups:  # no aperture groups, create them
+            model = automatic_aperture_grouping(
+                model, room_based=True, view_factor_or_orientation=False)
+        else:
+            raise ValueError('No aperture groups found in the model.')
+
+        unique_aperture_groups = get_unique_aperture_groups(model)
```
mostaphaRoudsari, line 977: Shouldn't we let the model that has aperture groups go through? Also I might be wrong but why would you need to recalculate the aperture groups twice?

```suggestion
        unique_aperture_groups = get_unique_aperture_groups(model)

        if unique_aperture_groups:
            # there are aperture groups in the model already
            pass
        elif not unique_aperture_groups and create_groups:  # no aperture groups, create them
            model = automatic_aperture_grouping(
                model, room_based=True, view_factor_or_orientation=False)
            unique_aperture_groups = get_unique_aperture_groups(model)
        else:
            raise ValueError(
                'No aperture groups found in the model. Either provide a model with aperture groups or '
                'use the --create-groups option to calculate the aperture groups as part of this command.'
            )

        
```

mostaphaRoudsari (commented):
The logic doesn't look correct to me but I'm missing the larger picture. I included my change suggestions.

honeybee_radiance/cli/multiphase.py
mikkelkp: You are right. I think I messed it up because I decided to change the code just before sending the PR. Your suggestion looks good to me.

PR #890, opened 2025-03-06
fix(translate): Add option to create grids

mikkelkp:


honeybee_radiance/cli/translate.py
```diff
@@ -149,6 +156,25 @@ def model_to_rad_folder(
...
+                unit_conversion = {
+                    'Meters': 1,
+                    'Millimeters': 0.001,
+                    'Feet': 0.3048,
+                    'Inches': 0.0254,
+                    'Centimeters': 0.01
+                }
+                grid_size = 24 * 0.0254 / unit_conversion.get(model.units, 1)
+                offset = 30 * 0.0254 / unit_conversion.get(model.units, 1)
+                wall_offset = 12 * 0.0254 / unit_conversion.get(model.units, 1)
```
mostaphaRoudsari, line 172: I suggest using metrics and IP values here based on the user input. I know that you're using the IP distances as a base, but it will generate non-round values in SI and people won't be happy about that. I believe there are round number recommendations in SI that are close to the IP values.

honeybee_radiance/cli/translate.py
```diff
@@ -149,6 +156,25 @@ def model_to_rad_folder(
...
+                for room in model.rooms:
+                    sensor_grids.append(room.properties.radiance.generate_sensor_grid(
+                        grid_size, offset=offset, wall_offset=wall_offset))
```
mostaphaRoudsari, line 175: We should most likely introduce a property to let the user tag which rooms should have sensor grids. This will need a change in the schema but I'm just leaving a comment here for now.

honeybee_radiance/cli/translate.py
```diff
@@ -159,6 +185,9 @@ def model_to_rad_folder(
             model, folder, config_file, minimal, views=view, grids=grid,
             full_match=full_match
         )
+
+        model.to_hbjson('output_model', '.')
```
mostaphaRoudsari, line 189: It's probably only a couple of seconds to re-write this model, but I suggest doing this only when there has been a change to the model. Otherwise, we can simply rename/copy the input file.

mostaphaRoudsari (commented):
Hi @mikkelkp, I added some small comments to the PR.

honeybee_radiance/cli/translate.py
mikkelkp: I changed this so it will simply copy the file.

honeybee_radiance/cli/translate.py
mikkelkp: I added round values for SI units.

